PR #6, opened 2018-10-28
Generated Code (first pass) + Entity

davedissian:


spatialos-sdk/Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "spatialos-sdk"
 version = "0.0.0"
-authors = ["Jamie Brynes <jamiebrynes7@gmail.com>"]
+authors = [
+    "Jamie Brynes <jamiebrynes7@gmail.com>",
+    "David Avedissian <git@dga.me.uk>"
```
jamiebrynes7: 🎉 

You know this means you are trapped now

src/worker/mod.rs
```diff
@@ -1,2 +1,5 @@
 pub mod core;
 pub(crate) mod internal;
+
+pub mod component;
+pub use self::component::*;
```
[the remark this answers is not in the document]
davedissian: Re-export everything from `component` module to `self` (i.e. ensuring the user doesn't need to write `component::` everywhere).

examples/project-example/generated_code.rs
```diff
@@ -0,0 +1,53 @@
...
+    x: Option<f32>
+}
+
+impl ComponentMetaclass for Example {
+    type Data = ExampleData;
+    type Update = ExampleUpdate;
+}
+
+impl ComponentUpdate<Example> for ExampleUpdate {
+    fn to_data(self) -> ExampleData {
```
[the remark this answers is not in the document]
davedissian: Removed it.

davedissian:
@jamiebrynes7 Ready for review. It may have pain points, and doesn't really integrate with the SDK yet, but it would be good to merge it now and improve it in subsequent PRs.

spatialos-sdk-tools/src/download_sdk/main.rs
```diff
@@ -20,7 +20,7 @@ fn main() {
 
     download_and_unpack(
         &SpatialPackageSource::WorkerSdk,
-        "c-static-x86_64-msvc_md-win32",
+        "c-static-x86_64-msvc_mt-win32",
```
davedissian: I switched to the Win32 lib which statically links the standard library. This avoids the need for everyone to install the MSVC 2015 redistributable all the time.

davedissian:
@randomPoison you may also want to take a look.

randomPoison:
Have you considered using serde for the serialization logic? It wouldn't add anything to serialization going to/from SpatialOS, but it would allow users to also serialize generated types to other formats, and ensures smoother interop with the wider Rust ecosystem.

spatialos-sdk-code-generator/src/bin.rs
```diff
@@ -0,0 +1,23 @@
...
+use spatialos_sdk_code_generator::generator;
+use spatialos_sdk_code_generator::schema_bundle;
+use std::env::args;
+use std::fs::File;
+use std::io::Read;
+use std::io::Write;
+
+pub fn main() {
+    let input_filename = args().nth(1).unwrap();
+    let output_filename = args().nth(2).unwrap();
```
randomPoison: Rather than parsing arguments manually, I recommend using a crate like [structopt](https://crates.io/crates/structopt) to handle the parsing logic for you. It'll make the code cleaner, and it'll provide a better UX by automatically generating help text and the like.

spatialos-sdk/src/worker/component.rs
```diff
@@ -1,35 +1,463 @@
...
+// A trait that's implemented by a component to convert to/from schema handle types.
+pub trait Component
+where
+    Self: std::marker::Sized,
+{
+    type Update;
+    type CommandRequest;
+    type CommandResponse;
+
+    fn component_id() -> ComponentId;
```
randomPoison: Would this make more sense as an associated const? i.e. `const ID: ComponentId;`

spatialos-sdk/examples/project-example/generated_code.rs
```diff
@@ -0,0 +1,1507 @@
...
+                output.field::<SchemaInt32>(1).add(input.value);
+                Ok(())
+            }
+        }
+
+        /* Components. */
+
+        #[derive(Debug)]
+        pub struct Example {
+            x: f32,
```
randomPoison: Looks like code generation is making all the fields of the generated structs private, should probably be public.

randomPoison:
btw I'm testing out this implementation by [trying to get entity creation working](https://github.com/jamiebrynes7/spatialos-sdk-rs/compare/generated_code...randomPoison:create-entity). I'm digging into the serialization logic now, I'll likely have more feedback as I get to understand it better.

spatialos-sdk/src/worker/component.rs
```diff
@@ -1,35 +1,463 @@
...
+
+    // Vtable helper functions.
+    unsafe fn handle_allocate(data: T) -> *mut raw::c_void {
+        // Allocate client handle and initialize with data.
+        let new_client_handle: *mut ClientHandle<T> =
+            mem::transmute(alloc::alloc(Layout::new::<ClientHandle<T>>()));
+        ptr::write(new_client_handle, ClientHandle::<T>::new(data));
+
+        // Return new pointer.
+        mem::transmute(new_client_handle)
```
randomPoison: I think the logic for `handle_allocate` could be much simpler. I think all it needs to be is

```
Arc::into_raw(Arc::new(data)) as *mut _
```

Using [`into_raw`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.into_raw) will give you the raw pointer, and will ensure that the counter isn't decremented until you use [`from_raw`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.from_raw) to restore it.

spatialos-sdk/src/worker/component.rs
```diff
@@ -1,35 +1,463 @@
...
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.component_vtables.len()
+    }
+}
+
+// An object which contains a reference counted instance to T.
+struct ClientHandle<T> {
+    data: Rc<T>,
```
randomPoison: `data` should probably an `Arc<T>` since the data will be sent to the networking thread and freed from there (if I'm understanding the serialization docs correctly).

spatialos-sdk/src/worker/component.rs
```diff
@@ -1,35 +1,463 @@
...
+    // When Drop() is called, data should reduce its reference count.
 }
 
-impl From<&Worker_ComponentUpdate> for ComponentUpdate {
-    fn from(update: &Worker_ComponentUpdate) -> Self {
-        ComponentUpdate {
-            component_id: update.component_id,
-            schema_type: update.schema_type,
+impl<T> ClientHandle<T> {
+    fn new(data: T) -> ClientHandle<T> {
```
randomPoison: Honestly, I don't even think you need `ClientHandle<T>` or a `new` function. You could just have `handle_allocate`, `handle_copy`, and `handle_free` as freestanding functions that manipulate an `Arc<T>` under the hood. At least, it doesn't seem to me like there's any value in exposing a `ClientHandle` type to the user when all you're going to do with it is cast it to a `*mut c_void` anyway.

spatialos-sdk/src/worker/component.rs
```diff
@@ -1,35 +1,463 @@
...
+        let client_handle_ptr: *mut ClientHandle<T> = mem::transmute(handle);
+
+        // Call drop() on pointer value.
+        client_handle_ptr.drop_in_place();
+
+        // Deallocate memory.
+        alloc::dealloc(
+            mem::transmute(client_handle_ptr),
+            Layout::new::<ClientHandle<T>>(),
+        );
```
randomPoison: I think this function could be simplified down to

```
let _ = Arc::<T>::from_raw(handle);
```

That will reconstruct the `Arc` from the raw pointer and then drop it, decrementing the counter (and thereby freeing the data if it was the last handle).

spatialos-sdk/src/worker/component.rs
randomPoison: Also, it's worth noting that you don't have to use `transmute` for pointer conversions, you can cast between pointer types.

spatialos-sdk/src/worker/component.rs
```diff
@@ -1,35 +1,463 @@
...
+        let client_handle_ptr: *mut ClientHandle<T> = mem::transmute(handle);
+
+        // Allocate client handle and initialize with copy of existing client handle
+        let new_client_handle: *mut ClientHandle<T> =
+            mem::transmute(alloc::alloc(Layout::new::<ClientHandle<T>>()));
+        let client_handle: &ClientHandle<T> = &*client_handle_ptr;
+        ptr::write(new_client_handle, client_handle.copy());
+
+        // Return new pointer.
+        mem::transmute(new_client_handle)
```
randomPoison: I think this function can be simplified down to:

```
let original = Arc::<T>::from_raw(handle);
let copy = original.clone();
mem::forget(original);
Arc::into_raw(copy) as *mut _
```

This will properly increment the counter for the `Arc` without

spatialos-sdk/src/worker/component.rs
davedissian: Yeah. It didn't feel right to have this extra layer of boxing/indirection, and I felt that it was actually redundant. However, I hadn't had the chance to think about it better after the initial prototyping.

spatialos-sdk/examples/project-example/generated_code.rs
davedissian: oops. you're right.

spatialos-sdk/src/worker/component.rs
davedissian: Yes, definitely. I didn't know consts in traits were also a thing.

spatialos-sdk/src/worker/component.rs
davedissian: Wow, that's nice. I wasn't aware of `into_raw` and `from_raw`. Does it essentially allocate space for refcount + T, and turn that memory block into a pointer to T (so the refcount is located at the memory address of T, minus a couple of bytes)?

spatialos-sdk/src/worker/component.rs
davedissian: Neat. Agreed.

spatialos-sdk/src/worker/component.rs
davedissian: Agree with this too.

spatialos-sdk/src/worker/component.rs
randomPoison: `Arc::new` allocates space for the refcount + `T`. `into_raw` simply calls `mem::forget` on the `Arc` (so that it doesn't get dropped and deallocate the data) and then returns a pointer to the data (you can see the implementation [here](https://doc.rust-lang.org/src/alloc/rc.rs.html#394-398)). When you call `from_raw`, it knows where the refcount is laid out relative to the data, so it can reconstruct the full `Arc` from just a pointer to the data (you can see the implementation [here](https://doc.rust-lang.org/src/alloc/rc.rs.html#429-444).

randomPoison:
So I think all my concerns have been addressed. There's lots more stuff to add, but I think it would make sense to get this merged into master soon so that I can update my other PRs accordingly and get them finished as well.

PR #9, opened 2018-10-31
Add Travis CI

jamiebrynes7:


jamiebrynes7:
Will need some insight from those that maintain `spatial` on how to set this up properly

PR #10, opened 2018-11-02
Add Entity Query Commands

jamiebrynes7:


jamiebrynes7:
Will want to write a nice test suite before merging this in - as the implementation is interesting

PR #31, opened 2018-12-01
Search for libs via env variable

### LINKED ISSUE #23: Search for SpatialOS libs through an environment variable
Currently, the location of the C API libraries is hard coded. Expand this to be read from an environment variable: `SPATIAL_RS_LIBS` or similar.

jamiebrynes7:
Changed where `spatialos-sdk-sys` looks for the C API libraries. 

Instead hardcoding it to look in `CARGO_MANIFEST_DIR/../dependencies` which would force us to ship libs when the crate is published, look in `SPATIAL_LIB_DIR` which is set by the user.

Fixes #23


jamiebrynes7:
@davedissian Want to get your opinion on something, should we set the `SPATIAL_LIB_DIR` automatically in `download_sdk` or is that considered bad manners?

Issue #24, opened 2018-11-22
Upgrade to SpatialOS SDK 13.5.1


jamiebrynes7:
Left to do: 

- [ ] Implement Alpha Locator connection
- [x] Implement KCP connection type
- [x] Connection status code changes

jamiebrynes7:
Left to do: 

- [ ] Implement Alpha Locator connection
- [ ] Implement KCP connection type
- [ ] Connection status code changes

PR #47, opened 2019-01-13
13.5.1 Upgrade - Connection Status changes

jamiebrynes7:
Addition of `get_connection_status_code` and `get_connection_status_detail`. Deprecated `is_connected`


jamiebrynes7:
Completes an item in #24

jamiebrynes7:
Looks like I'll need to properly resolve the dependencies problem before the linking will work.

PR #49, opened 2019-01-20
Add cargo-spatial-codegen tool

randomPoison:
This PR replaces the `setup.sh` helper script used in the project-example example with a new `setup` subcommand in the example application itself. This provides better cross-platform support for running the example, specifically it makes it easier to run the example on Windows.

Additional details:

* Added a new `setup.rs` module in project-example that replicates the logic in `setup.sh`.
* Added argument parsing for a new `setup` subcommand in `argument_parsing.rs`.
* Adds some additional logic in `main()` to handle the two different ways the example can be run (i.e. launching a worker or performing setup).
* Tweaked the logic for handling errors in the command line arguments to log directly to stderr instead of panicking).

Please let me know if there's anything I need to change!


randomPoison:
Looks like the travis build failed, but I'm unclear on what the error is? It seems like the environment variable `SPATIAL_OAUTH` is missing, any idea why my change would have caused that?

jamiebrynes7:
Ah that's because the PR is coming from a fork and the `SPATIAL_OAUTH` is a secret which Travis doesn't expose to PRs from forks. Will look into solutions for this 😄

jamiebrynes7:
Looks like I'll have to manually run the CI suite locally and override the failing check when we go to merge. Will take a look at the code later 👍

spatialos-sdk/examples/project-example/setup.rs
```diff
@@ -0,0 +1,81 @@
...
+            &tmp_path,
+            &CopyOptions {
+                overwrite: true,
+                ..CopyOptions::new()
+            },
+        )
+        .expect("Failed to copy contents of schema-compiler/proto");
+    }
+
+    // Run the schema compiler for each of the schema files in std-lib/improbable.
```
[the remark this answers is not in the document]
jamiebrynes7: A similar comment applies for the `protoc` compiler as well.

spatialos-sdk/examples/project-example/setup.rs
```diff
@@ -0,0 +1,81 @@
...
+    let schema_glob = spatial_lib_dir.join("std-lib/improbable/*.schema");
+    let schema_path_arg =
+        OsString::from("--schema_path=").tap(|arg| arg.push(&spatial_lib_dir.join("std-lib")));
```
jamiebrynes7: TIL about `tap` - pretty neat 👍

spatialos-sdk/examples/project-example/setup.rs
```diff
@@ -0,0 +1,81 @@
...
+    // dir path.
+    let schema_compiler_path = spatial_lib_dir.join("schema-compiler/schema_compiler");
+    let protoc_path = spatial_lib_dir.join("schema-compiler/protoc");
+
+    // Calculate the various output directories relative to `out_dir`.
+    let bin_path = out_dir.join("spatialos/schema/bin");
+    let tmp_path = out_dir.join("tmp");
+
+    // Create the output directories if they don't already exist.
+    fs::create_dir_all(&bin_path).expect("Failed to crate spatialos/schema/bin");
```
jamiebrynes7: Would we want to ensure that these directories are empty before we start working on them?

spatialos-sdk/examples/project-example/setup.rs
```diff
@@ -0,0 +1,81 @@
...
+use tap::*;
+
+pub fn do_setup(spatial_lib_dir: Option<PathBuf>, out_dir: PathBuf) {
+    let spatial_lib_dir = spatial_lib_dir
+        .or_else(|| std::env::var("SPATIAL_LIB_DIR").map(Into::into).ok())
+        .expect("--spatial_lib_dir argument must be specified or the SPATIAL_LIB_DIR environment variable must be set");
+
+    // Determine the paths the the schema compiler and protoc relative the the lib
+    // dir path.
+    let schema_compiler_path = spatial_lib_dir.join("schema-compiler/schema_compiler");
```
jamiebrynes7: Does omitting the `exe` extension on Windows still invoke the command properly?

randomPoison:
I agree that splitting the setup script into its own is a good idea. I've been trying to setup [my own project](https://github.com/randomPoison/online-fps/tree/spatialos) based on the example in this repo, and I'm finding that I need this setup script in order to compile the schema files. Splitting it off means that I can install it on its own, rather than having to copy-paste the script into my project.

spatialos-sdk/examples/project-example/setup.rs
randomPoison: Yeah, it's super useful! If you want to read more about it, this is the blog post that talks about it more in detail: https://myrrlyn.net/blog/misc/rust-flow-part-two

spatialos-sdk/examples/project-example/setup.rs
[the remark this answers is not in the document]
randomPoison: Ah, thank you for clarifying! I'm not super familiar with how glob patterns get expanded in bash, should be easy enough to restore the original behavior.

spatialos-sdk/examples/project-example/setup.rs
randomPoison: Which directories are you referring to? The output dir and the temp dir? I imagine it's fine if the output dir isn't empty (e.g. updating the generated files when schemas change), might be good to issue a warning if the temp dir already exists? Might be a better way to handle temp dirs in that case, though 🤔

randomPoison:
btw I'm marking this as WIP, I want to do a cleanup pass and make sure errors are being handled properly before this gets merged.

randomPoison:
I've cleaned up the error handling logic (i.e. not just panicking on every error) and removed the WIP marker.

spatialos-sdk/examples/project-example/setup.rs
jamiebrynes7: I'm unsure of the behaviour of the schema compiler if the output file already exists, I think it should just overwrite it, but can't guarantee of the top of my head.

Broadly it should be fine though :smile:

spatialos-sdk-tools/src/setup/main.rs
```diff
@@ -0,0 +1,106 @@
...
+#[structopt(name = "setup", about = "Perform setup for a Rust SpatialOS project.")]
+struct Opt {
+    /// The path to your local installation of the SpatialOS SDK. If not specified,
+    /// the SPATIAL_OS_DIR environment variable instead.
+    #[structopt(long = "spatial-lib-dir", short = "l", parse(from_os_str))]
+    spatial_lib_dir: Option<PathBuf>,
+
+    /// The path the schema directory for the project.
+    #[structopt(parse(from_os_str))]
+    schema_dir: PathBuf,
```
jamiebrynes7: Would prefer if this was named to `output_dir` rather than `schema_dir`

spatialos-sdk/examples/project-example/setup.rs
randomPoison: When I run it on my machine it appears to overwrite the existing file (the timestamp is updated, at least) so 🤷‍♀️

spatialos-sdk-tools/src/setup/main.rs
randomPoison: Done! Though in that case, does the current logic of appending `/bin` to `output_dir` make sense? I thought that was reasonable if the idea was that the user specifies their schema dir, we compile all their schemas, then output the compiled descriptor set to the `bin` dir inside their schema dir. If we're not explicitly asking for the schema dir, should I change it so that they specify the bin dir directly?

jamiebrynes7: Well you have have multiple input schema directories - which should compile down to one schema descriptor (not implemented yet, but a trivial thing to change). 

Ultimately you have some something like: 

```
cargo run --bin setup -- --schema-path=./my-user-path --schema-path=./my-other-schema-path --output-dir=./spatialos/schema
```

and the setup infers the std lib location from `SPATIAL_LIB_DIR` and globs any `schema-path` arguments

spatialos-sdk/examples/project-example/setup.rs
jamiebrynes7: Seems legit 💥

jamiebrynes7:
So running the command listed in the adjusted README gives me the following:

```bash
λ cargo run --bin setup -- spatialos-sdk/examples/project-example/spatialos/schema
   Compiling spatialos-sdk-tools v0.0.0 (C:\Workspace\personal\spatialos-sdk-rs\spatialos-sdk-tools)
    Finished dev [unoptimized + debuginfo] target(s) in 1.36s
     Running `target\debug\setup.exe spatialos-sdk/examples/project-example/spatialos/schema`
C:/Workspace\personal\spatialos-sdk-rs\dependencies\std-lib\improbable\standard_library.schema: Input file is not contained in any --schema_path. One --schema_path argument must be an exact prefix of the input file.
C:/Workspace\personal\spatialos-sdk-rs\dependencies\std-lib\improbable\vector3.schema: Input file is not contained in any --schema_path. One --schema_path argument must be an exact prefix of the input file.
```

Which looks like an error from the `schema-compiler`

randomPoison:
Hmmmmm, I was running into an issue similar to this before. It seems like `schema-compiler` requires `--schema-path` to be byte-for-byte identical with the leading part of any schemas you pass in. This can lead to it rejecting cases that should be valid. e.g. I ran into this on windows in cases where forward- and back-slashes got mixed up:

```
schema-compiler \
  --schema-path=C:\foo\bar\improbable/std-lib/schema \
  C:\foo\bar\improbable\std-lib\schema\whatever.schema
```

In that example `--schema-path` is logically a prefix of `whatever.schema`, however it mixes separators, which causes `schema-compiler` to reject the invocation.

You can actually see what is (I think) the cause in the example you posted:

```
C:/Workspace\personal\spatialos-sdk-rs\dependencies\std-lib\improbable\standard_library.schema
```

The first separator is `/` but the subsequent ones are `\`. I think the issue is in your `SPATIAL_LIB_DIR` variable, it probably defines the path with inconsistent separators.

Ultimately, this is an issue with `shema-compiler` in that it's rejecting perfectly valid paths. I can see if I can work around it for now by normalizing the paths before invoking the command.

randomPoison:
I was able to reproduce the issue by using the `--spatial-lib-dir` argument to manually specify a path with mixed separators. I've added path normalization logic that fixes the issue in the cases that I was able to test. The logic is a bit ugly, but tbh the "right" solution is to fix `schema_compiler` to handle paths better. In the meantime, I've added a HACK comment clarifying exactly why the functionality is needed.

As an aside, there's [a tracking issue](https://github.com/rust-lang/rust/issues/47402) for adding path normalization logic to std, but it hasn't landed on stable yet :sob:

randomPoison:
@jamiebrynes7 @davedissian I've updated this to add support for codegen in the setup script. `setup` can now run codegen, so there's only a single step users need to run in order to prepare their project (or the example).

I've also deleted the `generated_code.rs` from the example and set it up to be ignored by git. The generated file should probably not be checked into git, since it will be generated by the user as part of the setup process.

spatialos-sdk-tools/src/setup/main.rs
```diff
@@ -0,0 +1,201 @@
...
+    command.status().map_err(|_| "Failed to run protoc")?;
+
+    // Remove the temp directory once the setup process has finished.
+    fs::remove_dir_all(&tmp_path).map_err(|_| "Failed to remove temp dir")?;
+
+    // If the user specified the `--codegen` flag, run code generation with the bundle file
+    // that we just generated.
+    if let Some(codegen_out_path) = codegen {
+        let mut input_file =
+            File::open(&bundle_json_path).expect("Unable to open the test schema bundle.");
```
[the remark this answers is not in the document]
jamiebrynes7: And a few lines below as well

spatialos-sdk-tools/src/setup/main.rs
randomPoison: Support for multiple `--schema-path` flags has been added, btw.

README.md
```diff
@@ -7,24 +7,26 @@
...
 ## Running the Example Project
 
 To run the example project, you will need to:
 
-1. Run the code generator - `./spatialos-sdk/examples/project-example/codegen.sh`
-2. Build a release version of the RustWorker - `cargo build --example project-example --release`.
-3. In two terminals:
-   - Navigate to the `spatialos` directory and start spatial: `cd spatialos-sdk/examples/project-example/spatialos/ && spatial local launch`
-   - Run the example project worker - `cargo run --example project-example -- receptionist --worker_type RustWorker`
+1. Install the setup tool: `cargo install --path ./spatialos-sdk-tools --bin setup`
```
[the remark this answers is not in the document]
randomPoison: Ah, that's a good point. I think the naming issue is ultimately moot since I plan to replace all the independent tools with `cargo spatial`, but I don't know how long it'll be before I'm ready to make that PR so I think giving the setup tool a better name is a good idea in the meantime 👍

spatialos-sdk/examples/project-example/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Generated bindings to SpatialOS components, shouldn't be committed.
+generated_code.rs
```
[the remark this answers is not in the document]
randomPoison: I think that's a fair point, though I think it would make more sense to have small, controlled examples as part of the spatialos-sdk-code-generator crate, rather than using the example project for that purpose. That said, I'm happy to restore `generated_code.rs` if you'd prefer 🙂

spatialos-sdk-tools/src/setup/main.rs
```diff
@@ -0,0 +1,201 @@
...
+    let tmp_path = output_dir.join("tmp");
+    let bundle_json_path = output_dir.join("bundle.json");
+
+    // Create the output directories if they don't already exist.
+    fs::create_dir_all(&output_dir)
+        .map_err(|_| format!("Failed to create {}", output_dir.display()))?;
+    fs::create_dir_all(&tmp_path)
+        .map_err(|_| format!("Failed to create {}", tmp_path.display()))?;
+
+    // Copy the contents of the schema-compiler/proto dir into the temp dir.
```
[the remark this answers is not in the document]
randomPoison: Sweet! Do we need to invoke `protoc` at all? It was originally being used to generate the descriptor file, but it seems like `schema_compiler` is also capable of doing that (as I learned from #6).

spatialos-sdk-tools/src/setup/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    // Run the schema compiler for each of the schema files in std-lib/improbable.
+    let schema_path_arg = OsString::from("--schema_path=").tap(|arg| arg.push(&std_lib_path));
+    let proto_out_arg = OsString::from("--proto_out=").tap(|arg| arg.push(&tmp_path));
+    let bundle_json_arg =
+        OsString::from("--bundle_json_out=").tap(|arg| arg.push(&bundle_json_path));
+    let descriptor_out_arg = OsString::from("--descriptor_set_out=")
+        .tap(|arg| arg.push(normalize(output_dir.join("schema.descriptor"))));
+    let mut command = Command::new(&schema_compiler_path);
+    command
```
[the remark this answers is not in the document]
randomPoison: Unfortunately, I don't think I can combine these. On line 87 I conditionally add arguments to `command` (in a loop), which means I need the command bound to a variable. But if I try to bind the return value of `arg()` to a variable, I'll get a compiler error, e.g.

```
let mut command = Command::new(...)
    .arg(...)
    .arg(...);
```

gives the error:

```
error[E0716]: temporary value dropped while borrowed
  --> spatialos-sdk-tools\src/codegen/main.rs:79:23
   |
79 |     let mut command = Command::new(&schema_compiler_path)
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary which is freed while still in use
...
84 |         .arg("--load_all_schema_on_schema_path");
   |                                                 - temporary value is freed at the end of this statement
...
88 |         command.arg(&arg);
   |         ------- borrow later used here
   |
   = note: consider using a `let` binding to create a longer lived value
```

The only way to make this work is to bind the result of `Command::new()` to a variable, then call `arg()` on it in a chain.

---

As an aside, this is why I use the [tap](https://crates.io/crates/tap) crate and avoid writing methods that return `Self`. There's no really good way to do method chaining by returning `Self` in Rust, and tap is the only robust solution I've found.

spatialos-sdk-tools/src/setup/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    /// Display detailed log output
+    #[structopt(long, short)]
+    verbose: bool,
+
+    /// Perform code generation and put the output in the specified file
+    ///
+    /// If not specified, will not perform code generation.
+    #[structopt(long, short = "c", parse(from_os_str))]
+    codegen: Option<PathBuf>,
```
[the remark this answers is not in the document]
randomPoison: Hrmmm, so I originally had it as something like that, but changed it to `--codegen` because I wanted it to more clearly indicate that specifying that flag was also enabling codegen. I can change it back to `--codegen-out-path` if you think that's better, I don't have super strong feelings about it either way. Though, if you think `--codegen` is fine as the CLI flag and you just want the variable to be `codegen_out_path`, I can also do that 😁

spatialos-sdk-tools/src/setup/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    /// The path the output directory for the project
+    #[structopt(long, short, parse(from_os_str))]
+    output_dir: PathBuf,
```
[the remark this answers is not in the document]
randomPoison: `output_dir` is currently used for both `schema.descriptor` and `bundle.json`. Does it make sense to call it `descriptor_out_dir` when it's used for both? Or should those two be split into separate flags?

spatialos-sdk-tools/src/setup/main.rs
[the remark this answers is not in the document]
randomPoison: Ah, good catch! I fixed this, and also replaced the calls to `unwrap`/`expect` with proper error handling.

randomPoison:
Do we even have to list all the schema files when invoking `schema_compiler`? If I'm reading [the docs](https://docs.improbable.io/reference/13.5/shared/schema/introduction#using-the-schema-compiler-directly) correctly, the `--load_all_schema_on_schema_path` flag that we're passing in is already telling it to load all the schema files. If we don't need to manually pass in any files, that would allow us to simplify the code further.

randomPoison:
I'm going to close this PR in favor of working on a more robust `cargo-spatial` tool. A PR for that should be coming soon :3

PR #50, opened 2019-01-22
Implement PartialEq for RequestId and make id private

randomPoison:
The derived implementation of `PartialEq` for `RequestId<T>` required that `T: PartialEq`, which means that you couldn't directly compare request IDs for most request types. Fortunately, we don't actually need that trait bound since the only thing we're comparing is `id`, so we can manually implement `PartialEq` in order to directly compare two `RequestId` objects.

On a related note, I've also made `RequestId::id` private, since we no longer need to access it directly in order to compare `RequestId`s. This improves encapsulation, and has the added advantage making `RequestId` more type-safe, since you can now only compare two `RequestId`s if they have the same type parameter (i.e. you can't compare a `RequestId<Foo>` and a `RequestId<Bar>`).

### DIFF
--- a/spatialos-sdk/src/worker/mod.rs
+++ b/spatialos-sdk/src/worker/mod.rs
@@ -36,7 +36,7 @@ impl EntityId {
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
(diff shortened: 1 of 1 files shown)


randomPoison:
I opted to manually implement `Hash`, since the implementation is minimal and it seems like the more idiomatic approach than suppressing the lint.

jamiebrynes7:
Alright we are good to go!

PR #51, opened 2019-01-23
OpList and WorkerOp improvements and fixes

randomPoison:
I've made a number of improvements to `OpList` and `WorkerOp`, including some critical safety fixes that address some use-after-free issues that are currently present. Some variants of `WorkerOp` depend on data owned by the root `OpList`, but there were previously no lifetimes ensuring that a `WorkerOp` didn't outlive its `OpList`. I've added the appropriate lifetime markers to the various op types, and updated `OpList` and `WorkerOp` accordingly.

Additional minor improvements:

* `OpList` no longer allocates (previously done by creating a `Vec<WorkerOp>` internally), now it directly uses the from the underlying `Worker_OpList`.
* Added `OpList::iter` helper method in order to make `OpList` API more idiomatic.
* Added `OpList::len` added.

@davedissian If this is going to cause merge conflicts with the codegen changes, I'm fine to have this wait until after those are merged (assuming it's approved) and deal with the conflicts on my end.


spatialos-sdk/src/worker/commands.rs
```diff
@@ -6,31 +6,31 @@ use spatialos_sdk_sys::worker::{
 };
 
 // TODO: Wrap Schema_CommandRequest
-pub struct CommandRequest {
+pub struct CommandRequest<'a> {
     pub component_id: ComponentId,
-    pub schema_type: *mut Schema_CommandRequest,
+    pub schema_type: &'a Schema_CommandRequest,
```
davedissian: Note that this is going to conflict pretty hard with my branch :P

randomPoison: I figured as much. I'm down to address conflicts once your changes get merged in :+1:

spatialos-sdk/src/worker/component.rs
```diff
@@ -52,17 +52,23 @@ where
...
+    pub struct ComponentData<'a> {
         pub component_id: ComponentId,
         pub schema_type: SchemaComponentData,
-        pub user_handle: *mut Worker_ComponentDataHandle,
+        pub user_handle: *const Worker_ComponentDataHandle,
+
+        // NOTE: `user_handle` is borrowing data owned by the parent object, but it's a
+        // type-erased pointer that may be null, so we just mark that we're borrowing
+        // *something*.
+        pub _marker: PhantomData<&'a ()>,
```
jamiebrynes7: Did not know you could do this. Pretty neat

spatialos-sdk/src/worker/op.rs
```diff
@@ -16,40 +16,81 @@ use crate::worker::internal::utils::*;
...
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # use spatialos_sdk::worker::connection::*;
+    /// # let connection: WorkerConnection = unimplemented!();
+    /// for op in connection.get_op_list(0).iter() {
+    ///     // Process `op`.
+    /// }
+    /// ```
+    pub fn iter(&self) -> Iter<'_> {
```
jamiebrynes7: qq - What does `Iter<'_>` mean

spatialos-sdk/src/worker/op.rs
```diff
@@ -462,17 +502,17 @@ pub struct EntityQueryResponseOp {
...
+    component_data: component::internal::ComponentData<'a>,
 }
 
-impl AddComponentOp {
+impl<'a> AddComponentOp<'a> {
     pub fn get<C: Component>(&self) -> Option<&C> {
         if C::ID == self.component_data.component_id {
             // TODO: Deserialize schema_type if user_handle is null.
-            Some(unsafe { &*(self.component_data.user_handle as *const _) })
+            Some(unsafe { &*(self.component_data.user_handle as *const _ as *const _) })
```
jamiebrynes7: Is the double cast necessary?

spatialos-sdk/src/worker/component.rs
randomPoison: Yup! You actually have to use `PhantomData` if you have a generic parameter that you're not using in the type itself (as is the case here).

spatialos-sdk/src/worker/op.rs
randomPoison: It's new in the 2018 edition. If I recall correctly, you have to specify a lifetime when referencing a type that has a lifetime parameter, i.e. I couldn't do `fn iter(&self) -> Iter`, I'd have to do `fn iter<'a>(&'a self) -> Iter<'a>`. Doing `'_` says "the value I'm returning borrows some data, but I'm letting the compiler infer the lifetime of that data". The idea is that it makes it clear that the returned value borrows data, without all the boilerplate of manually writing out the lifetimes.

spatialos-sdk/src/worker/op.rs
randomPoison: Nope! It was briefly necessary in a previous iteration, but I changed things such that it's no longer necessary. Looks like I missed one when I was cleaning it up.

randomPoison:
Updated!

jamiebrynes7:
Same deal here - reopen PR on the repo directly and I'll reapprove 👍

PR #52, opened 2019-01-26
Vtable serialization improvements

randomPoison:


spatialos-sdk/examples/project-example/generated_code.rs
```diff
@@ -534,9 +534,9 @@ mod generated {
 
         #[derive(Debug)]
         pub struct Coordinates {
-            x: f64,
-            y: f64,
-            z: f64,
+            pub x: f64,
```
davedissian: I'll modify the code generation template to do this for everything.

randomPoison: Yeah, I hand't looked into that myself since I could manually fix it for now.

spatialos-sdk/examples/project-example/main.rs
```diff
@@ -58,19 +59,19 @@ fn logic_loop(c: &mut WorkerConnection) {
                         == generated_code::example::Example::component_id()
                     {
                         let component_data =
-                            component::get_component_data::<generated_code::example::Example>(
+                            unsafe { component::get_component_data::<generated_code::example::Example>(
```
davedissian: Can we move these unsafe blocks into the implementation of `get_component_data`?

spatialos-sdk/src/worker/entity.rs
```diff
@@ -0,0 +1,50 @@
+use crate::worker::component::{self, Component, ComponentId};
+use spatialos_sdk_sys::worker::Worker_ComponentData;
+use std::collections::HashSet;
+use std::ptr;
+
+#[derive(Debug, Default)]
+pub struct Entity {
+    component_ids: HashSet<ComponentId>,
+    pub(crate) component_data: Vec<Worker_ComponentData>,
```
davedissian: I'd probably make this a HashSet of ComponentId to the Rust representation of components, and then convert that to `Vec<Worker_ComponentData>` when passing it to the C API.

spatialos-sdk/src/worker/entity.rs
```diff
@@ -0,0 +1,50 @@
...
+            reserved: ptr::null_mut(),
+            component_id: C::component_id(),
+            schema_type: ptr::null_mut(),
+            user_handle: data_ptr as *mut _,
+        };
+
+        self.component_ids.insert(C::component_id());
+        self.component_data.push(component_data);
+        self.drops.push(drop_raw::<C>);
+    }
```
davedissian: Could you also add the functions `pub fn get<C: Component>(&self) -> &C` and `pub fn get_mut<C: Component>(&mut self) -> &mut C`?

spatialos-sdk/src/worker/connection.rs
```diff
@@ -260,10 +262,29 @@ impl Connection for WorkerConnection {
...
+            None => ptr::null(),
+        };
+        let entity_id = match entity_id {
+            Some(e) => &e.id,
+            None => ptr::null(),
+        };
+        let id = unsafe {
+            Worker_Connection_SendCreateEntityRequest(
+                self.connection_ptr,
+                entity.component_data.len() as _,
```
davedissian: Yeah, I would build this `component_data` array only when needed, rather than maintaining it in the `Entity` struct. This is the approach we took with the C++/C#/Java SDKs.

randomPoison: Okay, if that's the pattern that's worked in the past then I'll get it setup that way here as well 👍

spatialos-sdk/src/worker/entity.rs
randomPoison: I'm down to add `get`, but I'm not sure `get_mut` is safe to add. Once you pass the entity to the networking thread for serialization, it wouldn't be safe to get a mutable reference to the component data. I _think_ we can make it safe if we can ensure that the user always gives up ownership of the `Entity` when they send for serialization, but I'm not quite sure how to do that at the moment. That said, I should be able to safely add a `remove` method, which will allow the user to remove the component, modify it, and then add it back if they want.

randomPoison: Hmmmm, actually removing a component also presents a challenge if the data has already been sent to the networking thread. We can remove the component from the `HashMap` and drop the `Arc` handle in order to decrease the refcount, but we won't be able to return the component data to the user if the handle has already been copied. I can have it return a `Result<C, RemoveErr>` where `RemoveErr` is an enum with variants to distinguish between "no component was in the entity" and "component removed from entity but it's being used on another thread".

Do you have any specific use case in mind for this API? I'm not sure how we're expecting it to be used, so I can't tell what design for the API makes the most sense.

spatialos-sdk/src/worker/entity.rs
randomPoison: Done! I made it a `HashMap<ComponentId, Worker_ComponentData> because we need to type-erase the components anyway, so storing the raw pointer in `Worker_ComponentData::user_handle` works as well as any of the alternatives.

spatialos-sdk/src/worker/entity.rs
davedissian: I personally think of the `Entity` class as some kind of view into a specific Entity from the perspective of the worker. In theory, we may want to maintain this view by merging component updates into it, and that would require you to be able to mutate data currently stored within the `HashMap`.

I agree that it's definitely unsafe to modify components once they've been sent to the networking thread, so I would probably suggest instead that you make a copy of all the component data when converting the `HashMap` into `Vec<Worker_ComponentData>`, rather than relying on reference counting. This also avoids the component removal issue.

spatialos-sdk/examples/project-example/main.rs
davedissian: Good point. I'm working on deleting this functions and replacing them with an implementation of `TryFrom` instead, so will probably add the component ID verification to ensure it is safe.

spatialos-sdk/src/worker/entity.rs
```diff
@@ -0,0 +1,71 @@
+use crate::worker::component::{self, Component, ComponentId};
+use spatialos_sdk_sys::worker::Worker_ComponentData;
+use std::collections::HashMap;
+use std::ptr;
+
+#[derive(Debug)]
+struct ComponentData {
+    raw_data: Worker_ComponentData,
```
davedissian: As mentioned in my other comment, I'm quite against having the `Entity` having any knowledge of `Worker_ComponentData` and instead storing components inside a type erased `Box<Component>` instead, and copy the data later into an `Arc<Component>` later. Not sure how easy that is to accomplish if everything is typed erased though. Might need to rely on the component database passed into the connection object in some way.

spatialos-sdk/src/worker/op.rs
```diff
@@ -455,7 +455,23 @@ pub struct EntityQueryResponseOp {
...
+    component_data: component::internal::ComponentData,
+}
+
+impl AddComponentOp {
+    pub fn component_id(&self) -> ComponentId {
+        self.component_data.component_id
+    }
+
+    pub fn get<C: Component>(&self) -> Option<&C> {
+        if C::ID == self.component_id() {
```
davedissian: This is still unsafe if the user opts to do manual serialization (which admittedly is not really possible yet), because `user_handle` will be `null`. However, I am happy to merge this now and implement that case later.

spatialos-sdk/src/worker/entity.rs
davedissian: Another way to avoid the potential pitfalls of the networking thread (and avoid copying) would instead be to try and make sure there is no way for an Entity to be shared by both the application and the network threads at the same time. You can accomplish this pretty easily by only allowing the user to move an Entity object into `Connection::send_create_entity_request` (effectively consuming it).

spatialos-sdk/src/worker/entity.rs
randomPoison: Hrmmm, there are slight performance differences between the two approaches, but I'd expect them to be negligible. Really, storing the `Box<Component>` is likely more efficient because we aren't creating an `Arc<T>`, and dealing with the atomic refcount is likely the biggest performance difference between the two approaches.

I think we're imagining different purposes for `Entity`, though. I wasn't expecting it to be a general-purpose view into an entity's data, I was only planning on using it purely for the purpose of gathering data to send to the C API (hence why I'm inclined to just use `Worker_ComponentData` internally). I'm not yet sure what a more general-purpose component data API would look like at this point, I'll likely have to see what use cases I run into as I work more with SpatialOS.

spatialos-sdk/src/worker/op.rs
randomPoison: Ah, that's a good point. I'll add a null check to the pointer and a panic noting that manual serialization isn't implemented yet 👍

PR #55, opened 2019-01-30
OpList improvements

randomPoison:
Re-opened from #51.


randomPoison:
@jamiebrynes7 Are you cool with me merging once the PR has been approved and builds are passing? Or would you prefer that I let you do that?

Issue #56, opened 2019-02-02
Only generate schema bindings for the local crate


randomPoison:
The current method for code generation involves generating the binding code for all schema-defined types in a single Rust source file. While this approach is fairly simple, it has the major drawback of making it difficult (or impossible) to distribute schema files and schema-generated code as part of a crate. For example, the SDK can't provide any useful method or trait implementations for the `improbable::Position` component, since the component effectively doesn't exist until the end-user generates it.

An alternative approach would be for each crate to distribute the generated code associated with any schema types it defines. This would allow the crate to also provide useful inherent/trait impls for such types. The SDK would distribute the [standard schema library](https://docs.improbable.io/reference/13.5/shared/schema/standard-schema-library).

## Unanswered Questions

* [ ] How do crates distribute the schema files themselves (i.e. is it even possible to distribute non-Rust assets in a crate)?
* [ ] How can we have the schema compiler find schema files pulled in from third-party crates?

randomPoison:
One part of this might be to allow for code generation to happen as part of the `build.rs` script. Right now, users have to manually execute a command in order to perform code generation, which means if they first open their project without doing so, they'll end up with a bunch of compiler errors. If code generation happened as part of the build script, it would ensure that users could always run `cargo build` on their worker project without needing to remember any other manual steps.

jamiebrynes7:
For a local workflow, having the generated code in its own crate might be worth looking at.

This is a common pattern for C# workers where you have `csproj` that is just generated code (and that on build you run codegen) and another which is your worker code that depends on the `csproj`. This solves a few problems:

1. You get IDE features working out of the box for worker code. 
2. You avoid recompiling generated code whenever you change worker code. 

However it does incur some overhead - 

1. Change schema = need to remember to rebuild the code gen crate. 

---

How this translates into shipping schema with a crate, I'm not sure. Say a user publishes a crate with some game functionality "spatialos-overwatch-movement", it could have a dependency on "spatialos-overmatch-movement-gen-code" which contains the generated Rust code. 

The obvious limitation is that you cannot reuse the schema defined in the crate. You could ship schema with the generated code crate. I.e. -

```
spatialos-overmatch-movement-gen-code
---
 |- schema/
 |- src/
 |- Cargo.toml
```

But I'm not sure _how_ a user could access this schema.

randomPoison:
Giving it some more though, I think we should be able to distribute schema files via crates.io:

[...]

PR #58, opened 2019-02-03
Add cargo-spatial tool

randomPoison:
I've created a new [cargo subcommand](https://github.com/rust-lang/cargo/wiki/Third-party-cargo-subcommands) in order to act as a high(er)-level tool for managing and building Rust SpatialOS projects. Its goal is to be the one-stop-shop for managing, building, and running SpatialOS projects. For example, the `cargo spatial local launch` command performs all steps necessary to build workers and launch a local deployment, specifically:

* Run code generation
* Build workers
* Copy worker binaries to a build folder so that the entire `target` directory isn't zipped (which speeds up deployment)
* Run `spatial local launch`

## Current functionality

* Project-specific configuration can be defined in a `Spatial.toml` file. This file also marks the root of a SpatialOS project, similar to how `Cargo.toml` marks the root of a Cargo crate. Keeping a configuration file ensures that project-specific options don't have to be passed to the CLI on every invocation. Reasonable defaults are provided for all configuration options, such that a project following the expected configuration can have an empty `Spatial.toml` file (just to mark the root of the project) and still work as expected.
* `cargo spatial codegen` performs code generation, including schema compilation (this supersedes the previously-proposed `cargo-spatial-codegen` tool).
* `cargo spatial local launch` launches a local deployment, automatically performing all steps necessary to build workers first.
* `cargo generate component-id` generates a random component ID in the range of valid IDs. Useful when defining a new component type in a schema file.

## Future Extensions

The following are changes that I plan make immediately following this PR, or to add to the issue tracker for future implementation:

* Automatic SDK setup, downloading the Spatial CLI and the SDK. The goal would be to automate the entire setup process, such that users only need to install cargo-spatial and the rest is handled automatically.
* Merge the rest of spatialos-sdk-tools into cargo-spatial.
* Add commands for building and running workers, including connecting local workers to both local and cloud deployments.
* Add commands for deploying to the cloud.
* Add support for running a local deployment in the background and automatically rebuilding and re-connecting workers as the user makes changes (similar to how in the Unity GDK you can create a local deployment and then automatically run all your workers by hitting the play button in the editor).

## Additional Changes

In order to better demonstrate the expected workflow for SpatialOS projects, I've changed the example project from being an example in spatialos-sdk-rs to being a standalone crate in the repo. This was necessary because cargo-spatial isn't setup to build examples as workers, and it didn't seem sensible to add that functionality. In splitting the example project into its own crate, I was able to set it up to reflect a reasonable default workflow for worker development, and it can act as a better test project for dogfooding development of the Rust SDK and cargo-spatial.


README.md
```diff
@@ -20,10 +20,9 @@ If these steps complete successfully, the `spatialos-sdk` crate has been built a
...
-1. Run the code generator - `./spatialos-sdk/examples/project-example/codegen.sh`
-2. Build a release version of the RustWorker - `cargo build --example project-example --release`
-4. Navigate to `spatialos-sdk/examples/project-example/spatialos`
-5. Run `spatial local launch`
+1. Install cargo-spatial: `cargo install --path ./cargo-spatial`
+2. Navigate to `project-example`
+3. Run `cargo spatial local launch`
 
 This will start a local deployment of SpatialOS with one entity. The entity will have the `Example`
 component described in `spatialos-sdk/examples/project-example/spatialos/schema/example.schema`.
```
jamiebrynes7: This path needs to be adjusted

README.md
```diff
@@ -32,7 +31,7 @@ by opening the inspector (navigate to http://localhost:21000/inspector in your w
 want to manually launch another instance of the worker, run the following command:
 
 ```
-cargo run --example project-example -- --worker-id RustWorker999 --worker-type RustWorker receptionist
+cargo run -- --worker-id RustWorker999 --worker-type RustWorker receptionist
```
jamiebrynes7: Does this command actually work in the root directory?

cargo-spatial/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+edition = "2018"
+
+[dependencies]
+glob = "0.2.11"
+log = "0.4.1"
+rand = "0.6.5"
+simplelog = "0.5.3"
+serde = { version = "1.0.38", features = ["derive"] }
+serde_json = "1.0.38"
+spatialos-sdk-code-generator = { path = "../spatialos-sdk-code-generator" }
```
jamiebrynes7: What if we wanted to allow users to plugin their own code generators here? 

I can imagine a use case where a user doesn't want our default implementation at all and wants to write their own layer. Could this be something we can specify in the `Spatial.toml` ?

cargo-spatial/src/codegen.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::ffi::OsString;
+use std::fs::{self, File};
+use std::io::prelude::*;
+use std::path::*;
+use std::process::Command;
+use tap::*;
+
+/// Performs code generation for the project described by `config`.
+///
+/// Assumes that the current working directory is the root directory of the project,
```
jamiebrynes7: Should we check or assert against this at the start of the function?

cargo-spatial/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+pub mod codegen;
+pub mod config;
+pub mod local;
+pub mod opt;
+
+/// Generates a random, valid component ID.
```
jamiebrynes7: I'm not sure about the usefulness of this command. Generally, you would arrange your schema so that the ids have some semantic meaning (or at least I would). Such that grouped schema ids = grouped features.

cargo-spatial/src/local.rs
```diff
@@ -0,0 +1,53 @@
...
+    crate::codegen::run_codegen(&config).expect("Failed to run codegen");
+
+    // Use `cargo install` to build workers and copy the exectuables to the build
+    // directory.
+    //
+    // TODO: Manually copy the built executables instead of using `cargo install`.
+    // `cargo install` doesn't use the same build cache as normal builds, so it will
+    // sometimes result in unnecessary recompilation, which can slow down launch times.
+    if !launch.no_build {
+        let build_dir = PathBuf::from(&config.build_dir).join("debug");
```
jamiebrynes7: Should probably expose the ability to use `release` builds for local launch as well (although not default)

cargo-spatial/src/local.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::ffi::OsString;
+use std::path::*;
+use std::process;
+use tap::*;
+
+/// Prepares and launches a local deployment.
+///
+/// Before launching the deployment, this will first run code generation and build
+/// workers in the project. Assumes that the current working directory is the root
+/// directory of the project, i.e. the directory that has the `Spatial.toml` file.
```
jamiebrynes7: Again, should we check or assert against this?

cargo-spatial/src/local.rs
```diff
@@ -0,0 +1,53 @@
...
+pub fn launch(_opt: &Opt, _local: &Local, launch: &LocalLaunch) {
+    let config = Config::load().expect("Failed to load configuration");
```
jamiebrynes7: `"Failed to load configuration"` -> `"Failed to load configuration in current working directory"`

cargo-spatial/src/local.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    // Run codegen and such.
+    crate::codegen::run_codegen(&config).expect("Failed to run codegen");
```
jamiebrynes7: Does this swallow the errors from `codegen`? It would be preferable to bubble those up

cargo-spatial/src/local.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    // Run `spatial alpha local launch` with any user-specified flags.
+    let mut command = process::Command::new("spatial");
+    command.args(&["alpha", "local", "launch"]);
+    if let Some(launch_config) = &launch.launch_config {
+        let arg = OsString::from("--launch_config=").tap(|arg| arg.push(launch_config));
+        command.arg(arg);
+    }
+    command.status().expect("Failed to run spatial");
```
jamiebrynes7: If this errors - we should print `stdout/stderr` or direct the user to a log

cargo-spatial/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+        LevelFilter::Trace
+    } else {
+        LevelFilter::Warn
+    };
+    SimpleLogger::init(verbosity, Default::default()).expect("Failed to setup logger");
+
+    // Perform the operation selected by the user.
+    match &opt.command {
+        Command::Codegen => {
+            let config = Config::load().expect("Failed to load config");
```
jamiebrynes7: `"Failed to load configuration"` -> `"Failed to load configuration in current working directory"`

cargo-spatial/src/opt.rs
```diff
@@ -0,0 +1,78 @@
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+/// Build, run, and deploy SpatialOS workers written in Rust using Cargo.
```
jamiebrynes7: What happens if we have workers in other build systems? Do we want to support the whole ecosystem as well?

cargo-spatial/src/opt.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[structopt(name = "launch")]
+    Launch(LocalLaunch),
+}
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct LocalLaunch {
+    /// Don't build workers before launching the local deployment
+    #[structopt(long, short)]
+    pub no_build: bool,
```
jamiebrynes7: I'm torn on what the default behaviour of this should be. I.e. - build or no build. Personally, I'd lean toward no rebuild and leave it to the user to decide when to build (as its an expensive operation).

project-example/spatialos.json
```diff
@@ -0,0 +1,8 @@
+{
+  "configurationVersion": "0.1",
+  "projectName": "xxx",
+  "schemaDescriptor": "./build/schema/schema.descriptor",
+  "serverWorkers": [
```
jamiebrynes7: Please add the client worker back in as seen here - https://github.com/jamiebrynes7/spatialos-sdk-rs/blob/master/spatialos-sdk/examples/project-example/spatialos/spatialos.json

project-example/src/main.rs
```diff
@@ -15,10 +15,15 @@ use std::collections::BTreeMap;
...
+mod connection_handler;
+mod opt;
 
-use generated_code::example;
+// HACK: Include `generated.rs` directly rather than treating it like a regular
+// module file because it internally declares a module named `generated`, which
+// causes clippy to complain about nested modules with the same name. We'll
+// likely end up using `include!` if we switch to doing code generation in a
+// build script, but the generated code will be at a more reasonable place.
+include!("generated.rs");
```
jamiebrynes7: The large downside of this is that you no longer get IDE features about generated code from the IntelliJ Rust plugin (maybe RLS too?). 

I'd rather add explicit clippy exclusions than this

README.md
randomPoison: No, it need to be run in the `project-example` directory. I'll update the wording to clarify :+1:

cargo-spatial/Cargo.toml
randomPoison: I'm pretty confident that's something we'll want to support. For example, for the Amethyst GDK we'll need to be able to generate the appropriate trait implementations for the relevant Amethyst traits, as well as any other glue code necessary for syncing SpatialOS data with the local world state, so custom code generation will be necessary. That said, I'd rather wait to add that functionality when I have a more concrete use case (i.e. when I'm starting to put together custom code generation for Amethyst), if you're cool with that.

cargo-spatial/src/lib.rs
randomPoison: So that's something that I wanted to ask more about. In relation to #56 and the desire to be able to distribute schema files via crates.io (or really any package manager), there needs to be a way to guarantee that your component IDs are unique not just within your own schema files, but also with any component definitions that you pull in from a third party. Generally, the way this is handled is by generating random IDs and having the range of valid values be so huge as to make collision a statistical impossibility (i.e. UUIDs or Git commit hashes). I added the random generation functionality in order to assist this approach, though I worry that the fact that component IDs are limited to the range 100 to 536,870,911 (with a range of 10,000 values taken out of the middle) means that collisions are going to be still pretty likely.

How is this issue handled for other language SDKs? Do other languages even support pulling in schema definition files from a package manager?

cargo-spatial/src/codegen.rs
randomPoison: Yeah, I think that's a good idea. There's a bit more to this that we might want to support, i.e. traversing up the directory hierarchy to find a `Spatial.toml` file in order to support running `cargo spatial` commands from sub-directories of the project (which is something Cargo supports), but I think for now it's a good idea to add asserts around the invariants we currently have.

cargo-spatial/src/local.rs
randomPoison: I moved the logic for loading the config file into `main()`, since it already has better logic setup for performing error handling. I also improved the wording of the errors in `Config::load()` according to you suggestion 👍

cargo-spatial/src/local.rs
randomPoison: It doesn't swallow the errors, rather it will panic if there's an error in code generation, which isn't the best way to report errors to the user. I'll tweak it to pass the error along in a more idiomatic manner, though I think error handling/reporting in general could use another pass (it's the part that I was least sure how to handle well).

cargo-spatial/src/local.rs
randomPoison: You mean if `spatial local launch` returns with an error code? Where would be the right place to direct the user?

cargo-spatial/src/opt.rs
randomPoison: I'd put that firmly in "to be determine" territory. I really don't have a good sense at this point of what it would look like to pull in third-party workers written in Rust, let alone ones written in other languages. I think for now it makes sense to limit cargo-spatial to just handling Rust workers, and we can worry about handling more complex setups later.

cargo-spatial/src/opt.rs
randomPoison: I'm inclined to keep the workflow and idioms close to what cargo does for normal Rust projects, and `cargo run` will attempt to build your crate before running it. I really like this workflow since it completely removes the ambiguity of "did I remember to build my code before running it?". In fact, `cargo run` doesn't even have an option for _not_ building your code first (though I think it's reasonable to have that option here, at least for now while we're still figuring out what the workflow looks like). Other than the potential cost of building workers, is there any other reason why you think it would be good to default to not building them?

project-example/spatialos.json
randomPoison: Ah right, I had deleted that to see if it would even work without a client worker JSON specified and I forgot to put it back. Sorry about that!

Is there any reason to have the client worker JSON? There isn't a client worker in the example project as far as I can tell.

project-example/src/main.rs
randomPoison: I'm surprised IntelliJ Rust doesn't support this, the Rust compiler is able to correctly give compiler warnings/errors pointing into `generated.rs`, though I'm not sure how well RLS handles it (RLS is pretty flaky about completion in general, so I don't really notice if things aren't working correctly). I'll switch to a regular `mod` statement and add a clippy inclusion instead 👍

This issue might need to be revisited as part of #56, though. I'm pretty sure the `include!` macro is the only way to pull in code generated in a `build.rs` script, or at least that's the way recommended in the Rust docs.

cargo-spatial/src/local.rs
jamiebrynes7: In the old project layout it was `$(dir spatialos.json)/logs` 

Not sure where in FPL the logs are put 🤔

project-example/spatialos.json
jamiebrynes7: You can run a "client" worker by passing in the `--worker-type RustClient` flag to the project. I was planning on expanding this such that the behaviour did actually differ between the workers

cargo-spatial/src/lib.rs
jamiebrynes7: Its generally an adhoc thing - the GDKs are the first to really play with the idea of shipping schema as user code and we are still figuring it out.

cargo-spatial/src/opt.rs
davedissian: I would say that `cargo spatial` should have no opinion about packaging non-Rust workers. To me, `cargo spatial build` should just map a Rust project (with all code and dependencies) into an assembly zip.

project-example/src/main.rs
```diff
@@ -15,10 +15,15 @@ use std::collections::BTreeMap;
...
 use tap::*;
 
-mod generated_code;
-mod lib;
+mod connection_handler;
+mod opt;
 
-use generated_code::example;
+// HACK: Include `generated.rs` directly rather than treating it like a regular
+// module file because it internally declares a module named `generated`, which
```
davedissian: Btw, the internal module can be named whatever we want. Everything gets re-exported at the bottom anyway. It's just there to be a "root" module for all the inner modules to refer back to. We could potentially remove the nested module and use the file itself as the root module, not sure.

project-example/src/main.rs
randomPoison: I was able to confirm that RLS *does* still work when using `include!`, which is nifty.

project-example/src/main.rs
randomPoison: I think it makes sense to remove that root module, since it makes the structure of the generated file non-idiomatic, and gives the consuming code less control over the module structure. I don't want to do that as part of this PR, so I'll leave it as-is and it can be cleaned up later.

cargo-spatial/src/local.rs
randomPoison: Seems like it's the same directory. Though the command line output from the `spatial` command is already getting logged to the command line automatically (`Command` shares `stdout`/`stderr` with the parent process by default). Do you think it's necessary to direct the user to the logs dir in addition to that? Does the `spatial` command already do that if an error occurs?

randomPoison:
Oh interesting, by changing the generated code to be pulled in via `mod` instead of `include!`, the linting pass is now failing due to formatting issues. I'll see about disabling rustfmt for that module.

cargo-spatial/src/opt.rs
jamiebrynes7: For the workflow where you manually connect workers (~90% of my personal workflow) I wouldn't want the extra build.

Although in that case, I may just default back to `spatial` directly. 

This is probably fine for now, one extension to this would be to read the launch config and determine whether the rebuild is required. I think the ergonomics of this can be improved over time (particularly if you have workers in multiple langauges / build systems) - but its a step in the right direction.

cargo-spatial/src/local.rs
```diff
@@ -0,0 +1,70 @@
...
+            if config.local_build_profile == BuildProfile::Debug {
+                command.arg("--debug");
+            }
+
+            let status = command
+                .status()
+                .map_err(|err| format!("Failed to build worker binaries: {}", err))?;
+
+            if !status.success() {
+                return Err("An error occurred while building workers")?;
```
jamiebrynes7: Should we log the stdout / stderr of the cargo command to help users debug?

ci/install.sh
```diff
@@ -56,3 +56,9 @@ rm -rf "./temp"
 
 rustup component add rustfmt-preview
 rustup component add clippy-preview
+
+cargo install --path cargo-spatial --force
+
+cd project-example
```
jamiebrynes7: Use `pushd` and `popd` :wink:

cargo-spatial/src/local.rs
randomPoison: Already done! The stdout/stderr gets sent to the console by default, so the user can already see its output when they run `cargo spatial local launch`.

cargo-spatial/src/opt.rs
randomPoison: Hmmmm, I think it's worth having a separate discussion about the workflow you're used to using and how cargo-spatial can best support it. Since I'm new to using SpatialOS I don't have a good sense of what a "normal" workflow looks like, so I just setup enough functionality to support the basic workflow I was using (i.e. ensure that all workers have been built and then run `spatial local launch` for testing). I'll open an issue to discuss workflow ideas 🙂

PR #62, opened 2019-02-07
Implement Send for WorkerConnection

randomPoison:
I'd like to propose that we mark `WorkerConnection` as `Send`. As I understand it, it should be perfectly fine to send the connection between threads, and marking it `Send` allows users to stick the connection object in a `Mutex` and share it between threads in a minimally-safe way. I can't really think of a reason why `WorkerConnection` wouldn't be safe to mark as `Send`, though I've [started a thread on the forums](https://forums.improbable.io/t/thread-safety-of-worker-connection-object/5358) to discuss the specifics of thread safety with the connection object.

If there's any reason why we shouldn't mark `WorkerConnection` as `Send`, please let me know! It's currently blocking me from integrating the SDK into Amethyst, since the connection object is effectively pinned to the thread that creates it.

### DIFF
--- a/spatialos-sdk/examples/project-example/main.rs
+++ b/spatialos-sdk/examples/project-example/main.rs
@@ -84,7 +84,7 @@ fn logic_loop(c: &mut WorkerConnection) {
 fn exercise_connection_code_paths(c: &mut WorkerConnection) {
     c.send_log_message(LogLevel::Info, "main", "Connected successfully!", None);
     print_worker_attributes(&c);
-    check_for_flag(&c, "my-flag");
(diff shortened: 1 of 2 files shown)


randomPoison:
I got [a reply from Alistair on my forum post](https://forums.improbable.io/t/thread-safety-of-worker-connection-object/5358/2?u=randompoison), and the information he gave indicates to me that it should be safe to make `WorkerConnection` be `Sync` as well (as long as we enforce certain invariants about how the data is accessed). I'm going to close this PR and re-open once I've finished updating the code :3

PR #63, opened 2019-02-08
Make WorkerConnection Send/Sync

randomPoison:
Based on what [Alastair said on the forums](https://forums.improbable.io/t/thread-safety-of-worker-connection-object/5358/2?u=randompoison), the worker connection should be safe to share between threads if wrapped in a mutex. For our purposes, that means that `WorkerConnection` is already `Send`, which is all we need to wrap it in a `Mutex` and start sharing it between threads. Fortunately, with some minor adjustments, `WorkerConnection` can also be made `Sync`, which enables a wider variety threading models for managing the connection.

So long as we always access the underlying `*mut Worker_Connection` pointer through `&mut self`, Rust's ownership system will guarantee that only one thread has access to the connection at a time. So all we should need in order to safely implement `Sync` for `WorkerConnection` is to change any methods that access the raw pointer to take `&mut self`. I've made this change, and I've added the `MutPtr<T>` wrapper type to enforce statically that all methods accessing the pointer have to take `&mut self`.

Additionally, there are two pieces of connection data that *should* be safe to access concurrently from multiple threads: The worker ID and worker attributes. The C API currently doesn't guarantee that the methods for retrieving this data are safe to call concurrently, however the data is static for the lifetime of the connection, so we can retrieve it once at startup and cache it in `WorkerConnection`. Once the data is cached at the Rust level, we can use Rust's normal safety guarantees to allow us to access that data through a `&self` (and therefore allow it to be accessed concurrently from multiple threads).

### DIFF
--- a/spatialos-sdk/src/worker/connection.rs
+++ b/spatialos-sdk/src/worker/connection.rs
@@ -5,6 +5,7 @@ use futures::{Async, Future};
 
 use spatialos_sdk_sys::worker::*;
 
+use crate::ptr::MutPtr;
 use crate::worker::commands::*;
 use crate::worker::component;
 use crate::worker::component::internal::{CommandRequest, CommandResponse, ComponentUpdate};
@@ -15,7 +16,6 @@ use crate::worker::metrics::Metrics;
 use crate::worker::op::OpList;
 use crate::worker::parameters::{CommandParameters, ConnectionParameters};
 use crate::worker::{EntityId, InterestOverride, LogLevel, RequestId};
-use crate::ptr::MutPtr;
 
 #[derive(Copy, Clone, PartialOrd, PartialEq, Debug)]
 pub enum ConnectionStatusCode {
--- a/spatialos-sdk/src/worker/connection.rs
+++ b/spatialos-sdk/src/worker/connection.rs
@@ -490,6 +490,9 @@ impl Drop for WorkerConnection {
     }
 }
 
+// SAFE: The worker connection object is safe to send between threads.
+unsafe impl Send for WorkerConnection {}
+
 pub struct WorkerConnectionFuture {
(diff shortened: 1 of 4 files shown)


spatialos-sdk/src/worker/connection.rs
```diff
@@ -124,21 +125,34 @@ pub trait Connection {
...
-    #[deprecated(note = "Use get_connection_status_code == ConnectionStatusCode::Success instead.")]
-    fn is_connected(&self) -> bool;
+    fn get_connection_status_code(&mut self) -> ConnectionStatusCode;
 
-    fn get_connection_status_code(&self) -> ConnectionStatusCode;
-    fn get_connection_status_detail(&self) -> String;
-    fn get_worker_id(&self) -> &str;
-    fn get_worker_attributes(&self) -> Vec<String>;
-    fn get_worker_flag(&self, name: &str) -> Option<String>;
+    fn get_connection_status_detail(&mut self) -> String;
```
davedissian: Consider replacing the `get_connection_status_code`/`detail` pair with a single `get_connection_status` that returns a `struct ConnectionStatus(ConnectionStatusCode, String)`. The C API will probably do this eventually anyway.

spatialos-sdk/src/worker/connection.rs
```diff
@@ -147,9 +161,19 @@ impl WorkerConnection {
             let worker_id = Worker_Connection_GetWorkerId(connection_ptr);
             let cstr = CStr::from_ptr(worker_id);
 
+            let sdk_attr = Worker_Connection_GetWorkerAttributes(connection_ptr);
+            let attributes = ::std::slice::from_raw_parts(
```
jamiebrynes7: Nice touch 👍

PR #64, opened 2019-02-08
Implement command request/response methods

randomPoison:
* Implement `send_command_request`, `send_command_response`, and `send_command_error` on `WorkerConnection`.
* Have `send_command_error` return a `Result<(), NulError>` in case the user passes in a message that can't be safely converted to a C string.
* Remove `CommandParameters`, opting to pass the `allow_short_circuit` option directly to `send_command_request`.

### DIFF
--- a/spatialos-sdk/src/worker/commands.rs
+++ b/spatialos-sdk/src/worker/commands.rs
@@ -1,5 +1,57 @@
 use crate::worker::query::EntityQuery;
 use crate::worker::EntityId;
+use spatialos_sdk_sys::worker::Worker_CommandParameters;
+
+/// Additional parameters for sending command requests.
+///
+/// Additional parameters passed to [`Connection::send_command_request`]. Note that all
+/// parameters are kept private and the struct can only be initialized with default
+/// values in order to make it possible to add new parameters without a breaking change.
+///
+/// If you would like to use a method-chaining style when initializing the parameters,
+/// the [tap] crate is recommended. The examples below demonstrate this.
+///
+/// # Parameters
+///
+/// * `allow_short_circuit: false` allow the command to skip being routed through SpatialOS if
+///   the worker is sending a command to itself. This avoids a round trip, but
+///   SpatialOS cannot guarantee that the command will be fully delivered and executed.
+///   [See the docs for mroe information.][short-circuit]
+///
+/// # Examples
+///
+/// ```
+/// use spatialos_sdk::worker::commands::CommandParameters;
+/// use tap::*;
+///
+/// let params = CommandParameters::new()
+///     .tap(CommandParameters::allow_short_circuit);
+/// ```
+///
+/// [short-circuit]: https://docs.improbable.io/reference/13.5/shared/design/commands#properties
+/// [tap]: https://crates.io/crates/tap
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
+pub struct CommandParameters {
+    allow_short_circuit: bool,
+}
+
+impl CommandParameters {
+    /// Creates a new `CommandParameters` with default values for all parameters.
+    pub fn new() -> CommandParameters {
+        Default::default()
+    }
+
+    /// Sets the `allow_short_circuit` parameter to `true`.
+    pub fn allow_short_circuit(&mut self) {
+        self.allow_short_circuit = true;
+    }
+
+    pub(crate) fn to_worker_sdk(&self) -> Worker_CommandParameters {
+        Worker_CommandParameters {
+            allow_short_circuit: self.allow_short_circuit as u8,
+        }
+    }
+}
 
 #[derive(Debug)]
 pub struct IncomingCommandRequest {}
--- a/spatialos-sdk/src/worker/commands.rs
+++ b/spatialos-sdk/src/worker/commands.rs
@@ -16,7 +16,7 @@ use spatialos_sdk_sys::worker::Worker_CommandParameters;
 /// * `allow_short_circuit: false` allow the command to skip being routed through SpatialOS if
 ///   the worker is sending a command to itself. This avoids a round trip, but
 ///   SpatialOS cannot guarantee that the command will be fully delivered and executed.
-///   [See the docs for mroe information.][short-circuit]
+///   [See the docs for more information.][short-circuit]
 ///
 /// # Examples
 ///
--- a/spatialos-sdk/src/worker/connection.rs
+++ b/spatialos-sdk/src/worker/connection.rs
@@ -6,8 +6,8 @@ use futures::{Async, Future};
 use spatialos_sdk_sys::worker::*;
 
 use crate::worker::commands::*;
-use crate::worker::component;
 use crate::worker::component::internal::{CommandRequest, CommandResponse, ComponentUpdate};
+use crate::worker::component::{self, Component};
 use crate::worker::entity::Entity;
 use crate::worker::internal::utils::cstr_to_string;
 use crate::worker::locator::*;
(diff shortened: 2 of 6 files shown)


randomPoison:
@jamiebrynes7 I've opted to remove `CommandParameters` entirely since it has only one field. Is there a good reason I'm missing for keeping it around?

davedissian:
@randomPoison The reason `CommandParameters` (and `ComponentUpdateParameters`) is a struct, is because if we decide to add more parameters later, it is technically a breaking change, which we cannot introduce until 14.0 according to our release policy. Making extra parameters a struct allows us to add more parameters without a breaking change.

spatialos-sdk/src/worker/connection.rs
```diff
@@ -85,25 +82,25 @@ pub trait Connection {
...
 
-    fn send_command_request(
+    fn send_command_request<C: Component>(
         &mut self,
         entity_id: EntityId,
-        request: component::internal::CommandRequest,
+        request: C::CommandRequest,
         timeout_millis: Option<u32>,
-        command_parameters: CommandParameters,
+        allow_short_circuit: bool,
```
davedissian: I would rather not inline this, as I mentioned in the other comment.

spatialos-sdk/src/worker/connection.rs
```diff
@@ -328,30 +325,79 @@ impl Connection for WorkerConnection {
...
 
     fn send_command_failure(
         &mut self,
-        _request_id: RequestId<IncomingCommandRequest>,
-        _message: &str,
-    ) {
-        unimplemented!()
+        request_id: RequestId<IncomingCommandRequest>,
+        message: &str,
+    ) -> Result<(), NulError> {
```
davedissian: Why does this return a result? This operation can't fail.

spatialos-sdk/src/worker/connection.rs
```diff
@@ -328,30 +325,79 @@ impl Connection for WorkerConnection {
...
-        _timeout_millis: Option<u32>,
-        _command_parameters: CommandParameters,
+        entity_id: EntityId,
+        request: C::CommandRequest,
+        timeout_millis: Option<u32>,
+        allow_short_circuit: bool,
     ) -> RequestId<OutgoingCommandRequest> {
-        unimplemented!()
+        let timeout = match timeout_millis {
+            Some(c) => &c,
```
davedissian: I'm surprised this compiles. Neat.

spatialos-sdk/src/worker/connection.rs
```diff
@@ -328,30 +325,79 @@ impl Connection for WorkerConnection {
...
+        let params = Worker_CommandParameters {
+            allow_short_circuit: allow_short_circuit as _,
+        };
+
+        let request_id = unsafe {
+            Worker_Connection_SendCommandRequest(
+                self.connection_ptr,
+                entity_id.id,
+                &command_request,
+                0,
```
davedissian: This shouldn't be 0. The command ID corresponds to the index of the command in the component (first command with ID 1). This would involve making a change to the generated code to add some kind of `get_command_id()` method onto the request and response type. I'll probably do this now quickly (as I have time).

davedissian: I pushed a commit which does this.

spatialos-sdk/src/worker/connection.rs
jamiebrynes7: Agreed with @davedissian - would prefer if the struct was kept

spatialos-sdk/src/worker/connection.rs
jamiebrynes7: Its because of the error propagation of the `CString::new()`..

From https://doc.rust-lang.org/std/ffi/struct.CString.html#errors
> This function will return an error if the supplied bytes contain an internal 0 byte

spatialos-sdk/src/worker/connection.rs
jamiebrynes7: I think this is a change with NLL? Since the borrow of `c` is valid for the function scope its perfectly happy 😄

randomPoison: It's not NLL related, rather the `&u32` is getting coerced to a `*const u32` (which is one of the few implicit conversions that Rust allows), at which point the lifetimes get thrown away 😉

randomPoison:
I've restored `CommandParameters`, but I've made some tweaks based on the desire for being able to add parameters without a breaking change. Specifically, I've made the fields private and made it only possible to construct an instance of `CommandParameters` with default values. This ensures that adding new parameters won't be a breaking change. I've tried to clarify this in the documentation for `CommandParameters`, let me know if anything is unclear 🙂

I'll address the remaining feedback soon!

spatialos-sdk/src/worker/connection.rs
randomPoison: I'm a bit torn here on whether or not it makes sense to expose this error. On the one hand it seems like a bad API if it panics or arbitrarily rejects the user's message for having null bytes in the middle, on the other hand it seems like exposing the `NulError` leaks implementation details about the fact that the is getting converted into a `CString`. For now I went with what seemed like the least surprising option (not panicking unless the user chooses to do so), but I'm open to alternate approaches as well.

spatialos-sdk/src/worker/commands.rs
```diff
@@ -1,5 +1,57 @@
...
+///
+/// If you would like to use a method-chaining style when initializing the parameters,
+/// the [tap] crate is recommended. The examples below demonstrate this.
+///
+/// # Parameters
+///
+/// * `allow_short_circuit: false` allow the command to skip being routed through SpatialOS if
+///   the worker is sending a command to itself. This avoids a round trip, but
+///   SpatialOS cannot guarantee that the command will be fully delivered and executed.
+///   [See the docs for mroe information.][short-circuit]
```
jamiebrynes7: Nit: `mroe` -> `more`

spatialos-sdk/src/worker/connection.rs
```diff
@@ -1,21 +1,18 @@
...
 use crate::worker::locator::*;
 use crate::worker::metrics::Metrics;
 use crate::worker::op::OpList;
-use crate::worker::parameters::{CommandParameters, ConnectionParameters};
+use crate::worker::parameters::ConnectionParameters;
 use crate::worker::{EntityId, InterestOverride, LogLevel, RequestId};
+use futures::{Async, Future};
+use spatialos_sdk_sys::worker::*;
+use std::ffi::{CStr, CString, NulError};
+use std::ptr;
```
jamiebrynes7: Small nit: Judicious use of white space can organise imports nicely :smile:

spatialos-sdk/src/worker/connection.rs
jamiebrynes7: I'm happy to keep this here for now and we can revisit this as part of #68

spatialos-sdk/src/worker/connection.rs
randomPoison: Hrmmm, this might be a personal taste thing, but I think grouping all the imports together and letting rustfmt sort them is the better approach. rustfmt only seems to sort them if there's no spaces between them, so the only way to make sure imports are organized consistently across all files is to group them up and let rustfmt handle it.

My suggestion to get nicer organization would be to make more use of nested imports, e.g.:

```
use futures::{Async, Future};
use spatialos_sdk_sys::worker::*;
use std::{
    ffi::{CStr, CString, NulError},
    ptr,
};
```

That has the same affect (i.e. grouping related imports together) while still ensuring that rustfmt can fully format imports.

jamiebrynes7: That's a shame that `rustfmt` doesn't handle the whitespace well :confused: 

In which case I agree grouping the imports is a good compromise :)

spatialos-sdk/src/worker/connection.rs
```diff
@@ -1,18 +1,21 @@
...
+    commands::*,
+    component::{self, internal::ComponentUpdate, Component},
+    entity::Entity,
+    internal::utils::cstr_to_string,
+    locator::*,
+    metrics::Metrics,
+    op::OpList,
+    parameters::ConnectionParameters,
+    {EntityId, InterestOverride, LogLevel, RequestId},
+};
```
randomPoison: @jamiebrynes7 This is a good example of a place where I think the nested formatting style achieves the type of import grouping you're looking for. If this seems reasonable to you, I'll start using this going forward :)

PR #66, opened 2019-02-08
Removed inner module from generated code. Implemented component update sending.

davedissian:
Still doesn't seem to quite work yet. Not sure why.


spatialos-sdk/examples/project-example/main.rs
```diff
@@ -45,24 +45,83 @@ fn main() {
...
         for op in &ops {
-            println!("Received op: {:?}", op);
+            if let WorkerOp::Metrics(_) = op {
+                println!("Received metrics.");
+            } else {
+                println!("Received op: {:?}", op);
+            }
             match op {
+                WorkerOp::ReserveEntityIdsResponse(response) => {
+                    if let StatusCode::Success(response_data) = response.status_code {
```
jamiebrynes7: We could make this failure tolerant pretty easy 😉

spatialos-sdk/src/worker/parameters.rs
```diff
@@ -358,7 +358,7 @@ impl ProtocolLoggingParameters {
             CString::new(self.log_prefix.clone()).expect("Received 0 byte in supplied log prefix.");
 
         Worker_ProtocolLoggingParameters {
-            log_prefix: log_prefix_cstr.as_ptr(),
+            log_prefix: "protocol-log-".as_ptr() as _,//log_prefix_cstr.as_ptr(),
```
jamiebrynes7: As we discussed offline - we need to change this implementation

spatialos-sdk/examples/project-example/main.rs
```diff
@@ -45,24 +45,83 @@ fn main() {
...
+    let mut entity_id: Option<EntityId> = None;
 
     loop {
         let ops = c.get_op_list(0);
 
+        // Move entity (if it exists).
+        if let Some(ref entity_id) = entity_id {
+            // Counter == 20 is one full circle.
+            let angle = (counter as f64) * (f64::consts::PI * 2.0 / 20.0);
+            c.send_component_update::<improbable::Position>(
```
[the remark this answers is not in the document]
davedissian: Yes it is explicitly required.

spatialos-sdk/src/worker/parameters.rs
davedissian: Yep. Will fix this properly later.

spatialos-sdk/examples/project-example/main.rs
```diff
@@ -45,24 +45,83 @@ fn main() {
...
+                entity_id.clone(),
+                improbable::PositionUpdate {
```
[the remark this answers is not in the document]
davedissian: Can you open a ticket?

spatialos-sdk/examples/project-example/main.rs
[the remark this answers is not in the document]
jamiebrynes7: Hmmm I wonder if that's just a failure of the compiler to infer `C` given `C::Update` 😕

spatialos-sdk/examples/project-example/main.rs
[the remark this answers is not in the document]
jamiebrynes7: See: #67

spatialos-sdk/examples/project-example/main.rs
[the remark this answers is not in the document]
davedissian: So I tried `improbable::Position::Update` (the trait alias) instead of `improbable::PositionUpdate` (the concrete type), but Rust wanted me to opt in explicitly to that trait by writing `<improbable::Position as Trait>::Update` or similar.

spatialos-sdk/examples/project-example/main.rs
```diff
@@ -45,24 +45,83 @@ fn main() {
...
-            println!("Received op: {:?}", op);
+            if let WorkerOp::Metrics(_) = op {
+                println!("Received metrics.");
+            } else {
+                println!("Received op: {:?}", op);
+            }
             match op {
+                WorkerOp::ReserveEntityIdsResponse(response) => {
+                    if let StatusCode::Success(response_data) = response.status_code {
+                        let mut entity = Entity::new();
```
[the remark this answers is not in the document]
randomPoison: I didn't factor this out into a separate method, but I did move this out of the `match` block to it's own section of the function, which I think makes the setup logic a bit cleaner. I'm hesitant to pull the entity creation logic out into its own function since it's only used in one place, but if you'd still prefer it to be separated then I can go ahead and do that 🙂

spatialos-sdk/examples/project-example/main.rs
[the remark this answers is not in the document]
randomPoison: This has been a known issue for a while: https://github.com/rust-lang/rust/issues/38078. Unfortunately the best we can do about it for now is to explicitly specify the trait as @davedissian has already done 😞

spatialos-sdk/examples/project-example/main.rs
randomPoison: Could you clarify what you mean here?

spatialos-sdk/src/worker/parameters.rs
randomPoison: I've addressed the safety issue here by having `ProtocolLoggingParameters` own the `CString` directly. In order to avoid leaking implementation details (i.e. that it's using `CString` under the hood), I've made all the fields private and added setters. I've also added a note in the doc comment about allowing us to add parameters without breaking changes (similar to `CommandParameters`), though if that's not appropriate here let me know.

I've also opted to use `expect()` on the result of the `CString` conversion instead of returning a `Result<_, NulError>`. After giving it more thought, I don't think it makes a ton of sense to expose that error directly. While it's _technically_ possible for a Rust string to contain a nul byte in the middle, it's both incredibly unlikely to happen in practice, and it's likely indicative of an error in the user's code.

randomPoison:
With @davedissian permission, I've gone ahead and finished up this PR. I've setup the example project to demonstrate component updates by adding a `Rotate` component that causes the entities to rotate in a circle a specified center point with a specified radius. I've also set this up so that it demonstrates changes in authority, such that a given worker will only try to update entities that it has authority over. The gif below shows entities moving smoothly while moving through the areas of authority of four different workers:

spatialos-rotate

The following gif also shows multiple entities moving in and out of a single worker's area of authority/intersted:

spatialos-rotate-authority

@jamiebrynes7 @davedissian I'm going to give the code another review pass myself, but it should be ready for you to re-review at this point.

project-example/schema/example.schema
```diff
@@ -8,4 +8,22 @@ component Example {
...
+    id = 1001;
+
+    /// The current angle (in radians) offset from the base.
+    double angle = 1;
+
+    /// The radius of the rotation.
+    double radius = 5;
+
+    /// The center point of the rotation.
+    // TODO: Use `improbable.Vector3d` here. Currently can't be used due to issues
```
jamiebrynes7: Can you open an issue for this? 😄

project-example/src/main.rs
```diff
@@ -44,40 +48,188 @@ fn main() {
...
+                // component.
                 WorkerOp::ComponentUpdate(update) => match update.component_id {
-                    example::Example::ID => {
-                        let component_update = update.get::<example::Example>();
-                        println!("Received Example update: {:?}", component_update)
+                    example::Rotate::ID => {
+                        let component_update = update.get::<example::Rotate>().unwrap();
+                        let state = world.get_mut(&update.entity_id).unwrap();
+                        let rotate = state.rotate.as_mut().unwrap();
+                        rotate.merge(component_update.clone());
```
jamiebrynes7: Should the `merge` method take the update by reference rather than value?

spatialos-sdk/src/worker/component.rs
```diff
@@ -53,6 +53,60 @@ where
...
+/// all parameters are kept private and the struct can only be initialized with
+/// default values in order to make it possible to add new parameters without a
+/// breaking change.
+///
+/// If you would like to use a method-chaining style when initializing the parameters,
+/// the [tap] crate is recommended. The examples below demonstrate this.
+///
+/// # Parameters
+///
+/// * `loopback: false` allow the update to be sent back to the worker without waiting
```
jamiebrynes7: Given the implementation: `loopback: true` allows for the short circuiting behaviour. 😉

spatialos-sdk/src/worker/component.rs
```diff
@@ -53,6 +53,60 @@ where
...
+///   to be routed through SpatialOS. This allows the worker to receive the update op
+///   faster, but risks that the worker won't know if the update actually failed. [See
```
jamiebrynes7: Technically, the worker doesn't know if the update failed either way! The difference is you could hit a race condition with authority changing mid tick I think

spatialos-sdk/examples/project-example/main.rs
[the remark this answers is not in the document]
jamiebrynes7: Yup it looks fine now - it was the nesting which I disliked 👍

spatialos-sdk/examples/project-example/main.rs
jamiebrynes7: Commands can fail and you can effectively retry the command with a bit of book keeping

spatialos-sdk/src/worker/component.rs
randomPoison: Ah sorry, the wording there is confusing. I only meant to indicate that the default value is `false`. I'll try to tweak the wording to make it more clear!

spatialos-sdk/src/worker/component.rs
randomPoison: Ah, good point. I'll update the wording to more accurately reflect the caveats with short-circuiting.

project-example/src/main.rs
randomPoison: I'm fine with it taking the update by value. If merge were to take the update object by reference, then it would need to copy the individual fields in order to apply them to the component. Since it ultimately needs the component data by value, it makes more sense to take the update by value, that way the calling code can determine if the value needs to be cloned. This seems to be the more common convention in Rust (i.e. let the calling code decide whether or not the clone the data).

project-example/schema/example.schema
randomPoison: Added #72 to track this :+1:

spatialos-sdk/examples/project-example/main.rs
randomPoison: Ah, good to know! I ended up removing the logic that depends on the request succeeding (i.e. it now unconditionally creates the entity without first reserving an ID), but I can still add the logic for repeating the request if you'd like 🙂

spatialos-sdk/src/worker/component.rs
```diff
@@ -94,6 +94,11 @@ impl UpdateParameters {
         Default::default()
     }
 
+    /// Enables loopback short-circuiting for the component update message.
+    ///
+    /// [See the docs for more information.][short-circuit]
+    ///
+    /// [short-circuit]: https://docs.improbable.io/reference/13.5/shared/design/commands#properties
```
jamiebrynes7: This link isn't quite correct. I don't think we have a link for this one yet since the C API doesn't have full docs and this isn't exposed in the other SDKs yet.

Maybe just put a TODO here to put a link when its available?

randomPoison: Done!

PR #73, opened 2019-02-20
SpatialOS SDK 13.6 Upgrade

### LINKED ISSUE #69: Upgrade to SpatialOS SDK 13.6.0
After #24, need to then upgrade to `13.6.0`! Has some minor changes to KCP and some schema compiler semantic changes.

jamiebrynes7:
This PR upgrades the underlying SDK to 13.6.0. This has the following changes: 

- Bindings updated (lots of `u32` changed to `i32`, need to look into this).
- KCP is now stable and has some small changes to the configuration
- The structure of some of the dev auth flow structs has changed.
- Removed the need for schema file globbing 

Fixes #69


cargo-spatial/src/codegen.rs
```diff
@@ -63,14 +63,6 @@ pub fn run_codegen(config: &Config) -> Result<(), Box<dyn std::error::Error>> {
         command.arg(&arg);
     }
 
-    // Add all schema files in the std lib.
-    add_schemas(&std_lib_path.join("improbable"), &mut command);
```
randomPoison: 🎉

jamiebrynes7:
Did some digging why the `u32` and `i32` change. According to https://github.com/rust-lang/rust-bindgen/issues/1096#issuecomment-365618652 it looks like its due to Clang differences.

In any case, it doesn't make a difference _here_

PR #78, opened 2019-03-01
Use inventory for automatic component registration 

### LINKED ISSUE #57: Automatic component registration
Currently users have to manually build the `ComponentDatabase` in order to enable automatic serialization. This means that they have to remember to add every single component in the standard schema library, their project, and any types from third-party libraries they use. This process is tedious and error-prone, and would benefit from an automated solution. There are two possible approaches I can think of:

First, have the code generation process generate a helper function that builds a `ComponentDatabase` with all the component types. Since code generation currently knows about all component types in the project, it should be relatively trivial to do this. However, #56 proposes that we change that, so we'll have to be cognizant of how these two proposals interact.

The other approach is to use the [inventory](https://crates.io/crates/inventory) crate to gather up all component types defined in all compiled crates (including third-party crates) and build a `ComponentDatabase` from them. This may interact better with #56 if we can't build the component database directly as part of code generation.

### LINKED ISSUE #71: Update Project for CLI changes
Some recent changes to the Spatial CLI included breaking changes for FPL. We will need to update the project to account for these.

See https://github.com/spatialos/FlexibleProjectExample/commit/cf24596174b1f4ba2b759cf4df9f85f4fead3937 for an example of what changes were made.

jamiebrynes7:
This PR introduces the use of the inventory crate for automatic component registration. 

- Added `WrappedVTable` as you must own the type that you call `inventory::collect!()` on. 
	- Could look into whether a typedef is sufficient for this.
- Driveby fix on the worker configs (FPL changes) 

Fixes #57 
Fixes #71 (most of the work appeared to have been done for this..?)

### DIFF
--- a/project-example/src/connection_handler.rs
+++ b/project-example/src/connection_handler.rs
@@ -30,7 +30,7 @@ pub fn get_connection(opt: Opt) -> Result<WorkerConnection, String> {
             let params = ConnectionParameters::new(worker_type)
                 .using_tcp()
                 .using_external_ip(connect_with_external_ip)
-                .generate_component_vtable();
+                .enable_internal_serialization();
             WorkerConnection::connect_receptionist_async(
                 &worker_id,
                 &host.unwrap_or_else(|| "127.0.0.1".into()),
@@ -53,7 +53,7 @@ pub fn get_connection(opt: Opt) -> Result<WorkerConnection, String> {
                 &ConnectionParameters::new(worker_type)
                     .using_tcp()
                     .using_external_ip(true)
-                    .generate_component_vtable(),
+                    .enable_internal_serialization(),
                 queue_status_callback,
             )
         }
@@ -97,7 +97,7 @@ pub fn get_connection(opt: Opt) -> Result<WorkerConnection, String> {
                 &ConnectionParameters::new(worker_type)
                     .using_tcp()
                     .using_external_ip(true)
-                    .generate_component_vtable(),
+                    .enable_internal_serialization(),
             )
         }
     };
--- a/project-example/src/connection_handler.rs
+++ b/project-example/src/connection_handler.rs
@@ -2,7 +2,6 @@ use crate::{Command, Opt};
 use futures::{Async, Future};
 use spatialos_sdk::worker::{
     alpha::{self, LoginTokensRequest, PlayerIdentityCredentials, PlayerIdentityTokenRequest},
-    component::ComponentDatabase,
     connection::{WorkerConnection, WorkerConnectionFuture},
     constants::{LOCATOR_HOSTNAME, LOCATOR_PORT, RECEPTIONIST_PORT},
     locator::{Locator, LocatorCredentials, LocatorParameters},
@@ -13,7 +12,7 @@ use uuid::Uuid;
 const POLL_NUM_ATTEMPTS: u32 = 5;
 const POLL_TIME_BETWEEN_ATTEMPTS_MILLIS: u64 = 3000;
 
-pub fn get_connection(opt: Opt, components: ComponentDatabase) -> Result<WorkerConnection, String> {
+pub fn get_connection(opt: Opt) -> Result<WorkerConnection, String> {
     let Opt {
         worker_type,
         worker_id,
@@ -28,9 +27,10 @@ pub fn get_connection(opt: Opt, components: ComponentDatabase) -> Result<WorkerC
             port,
             connect_with_external_ip,
         } => {
-            let params = ConnectionParameters::new(worker_type, components)
+            let params = ConnectionParameters::new(worker_type)
                 .using_tcp()
-                .using_external_ip(connect_with_external_ip);
+                .using_external_ip(connect_with_external_ip)
+                .generate_component_vtable();
             WorkerConnection::connect_receptionist_async(
                 &worker_id,
                 &host.unwrap_or_else(|| "127.0.0.1".into()),
@@ -50,9 +50,10 @@ pub fn get_connection(opt: Opt, components: ComponentDatabase) -> Result<WorkerC
             WorkerConnection::connect_locator_async(
                 &locator,
                 &deployment,
-                &ConnectionParameters::new(worker_type, components)
+                &ConnectionParameters::new(worker_type)
                     .using_tcp()
-                    .using_external_ip(true),
+                    .using_external_ip(true)
+                    .generate_component_vtable(),
                 queue_status_callback,
             )
         }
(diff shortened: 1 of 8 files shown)


project-example/src/main.rs
```diff
@@ -24,16 +24,8 @@ mod generated;
...
-    let components = ComponentDatabase::new()
-        .add_component::<example::Example>()
-        .add_component::<example::Rotate>()
-        .add_component::<improbable::EntityAcl>()
-        .add_component::<improbable::Persistence>()
-        .add_component::<improbable::Metadata>()
-        .add_component::<improbable::Interest>()
-        .add_component::<improbable::Position>();
+
+    let components = ComponentDatabase::new();
```
randomPoison: 🎉✨💖✨🎉

spatialos-sdk/src/worker/component.rs
```diff
@@ -271,8 +274,20 @@ pub(crate) unsafe fn handle_copy<T>(handle: *mut raw::c_void) -> *mut raw::c_voi
     Arc::into_raw(copy) as *mut _
 }
 
+pub struct WrappedVTable {
+    vtable: Worker_ComponentVtable
+}
+
+impl WrappedVTable {
+    pub fn new(vtable: Worker_ComponentVtable) -> Self {
```
randomPoison: Rather than taking the C type as a parameter, I'd suggest having `new` take the component type as a type parameter and then call `create_component_vtable` internally. This simplifies the public interface, and allows us to keep `create_component_vtable` a private implementation detail.

spatialos-sdk/src/worker/component.rs
```diff
@@ -271,8 +274,20 @@ pub(crate) unsafe fn handle_copy<T>(handle: *mut raw::c_void) -> *mut raw::c_voi
...
+        WrappedVTable {
+            vtable
+        }
+    }
+}
+
 // Vtable implementation functions.
-fn create_component_vtable<C: Component>() -> Worker_ComponentVtable {
+pub fn create_component_vtable<C: Component>() -> Worker_ComponentVtable {
```
randomPoison: As noted in the other comment, I think it would be better to keep this private and have `VTable::new` call it internally.

spatialos-sdk/src/worker/component.rs
```diff
@@ -237,14 +239,15 @@ pub struct ComponentDatabase {
 
 impl ComponentDatabase {
     pub fn new() -> Self {
+        let mut vtables = Vec::new();
+        for table in inventory::iter::<WrappedVTable> {
```
davedissian: How does this work? Is it reflection done at runtime, or is the inventory something generated at compile time? This is really really neat.

spatialos-sdk/src/worker/component.rs
davedissian: +1

spatialos-sdk/src/worker/component.rs
```diff
@@ -271,8 +274,20 @@ pub(crate) unsafe fn handle_copy<T>(handle: *mut raw::c_void) -> *mut raw::c_voi
     Arc::into_raw(copy) as *mut _
 }
 
+pub struct WrappedVTable {
```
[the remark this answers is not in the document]
davedissian: Agreed.

spatialos-sdk/src/worker/component.rs
davedissian: Agreed. Could probably just roll `create_component_vtable` directly into `new` as well.

spatialos-sdk/src/worker/component.rs
```diff
@@ -237,14 +239,15 @@ pub struct ComponentDatabase {
 
 impl ComponentDatabase {
```
davedissian: I think the ComponentDatabase type is completely redundant at this point. Might be worth moving the code that generates a vtable array into a private free standing function that is called by the connection/snapshot `new` functions when generating a vtable to pass to the C API.

project-example/src/main.rs
davedissian: Lets delete this line too, and just roll it into the connection object.

spatialos-sdk/src/worker/component.rs
jamiebrynes7: From https://github.com/dtolnay/inventory: 

> Inventory is built on the ctor crate which provides module initialization functions for Rust similar to __attribute__((constructor)) in C. Each call to inventory::submit! produces a shim that evaluates the given expression and registers it into a registry of its corresponding type. This registration happens dynamically as part of life-before-main for statically linked elements. Elements brought in by a dynamically loaded library are registered at the time that dlopen occurs.

spatialos-sdk/src/worker/component.rs
jamiebrynes7: Yeah that makes sense 👍

spatialos-sdk/src/worker/component.rs
jamiebrynes7: Yeah I was having a similar thought - was looking into if we can just expose a single static `ComponentDatabase` instance that any type that requires it could reference, but `lazy_static` and `* c_void` are not compatible. 

Equally, the cost of alloc'ing a vtable each time its required isn't likely to be particularly high

spatialos-sdk/src/worker/component.rs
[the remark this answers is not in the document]
jamiebrynes7: Yup - good call

spatialos-sdk/src/worker/component.rs
jamiebrynes7: That's a good shout

spatialos-sdk/src/worker/component.rs
davedissian: Yeah, a vtable is just a 16 function pointers for each component

randomPoison: We could probably switch to having an optional flag on `WorkerConnection` like `use_vtable_serialization` that enables vtable registration, and then gather the vtables internally.

jamiebrynes7:
Alright @davedissian, @randomPoison I think I've addressed the comments above: 

- Renamed `WrappedVTable` to `VTable`.
- Pulled `create_component_vtable<C: Component>()` into `VTable`.
- Made `ComponentDatabase` internal to the crate
- Added `generate_component_vtable()` method to `ConnectionParameters` to opt into the vtable (on second thought - it may be worth having this behaviour be default and have a `remove_vtables()` method if a user wants to use the schema types directly.
- Re-exported `inventory` in `worker::component` so the consumer of the generated code doesn't need to add `inventory` to their Cargo.toml and thus leak implementation details

project-example/src/connection_handler.rs
```diff
@@ -93,9 +94,10 @@ pub fn get_connection(opt: Opt, components: ComponentDatabase) -> Result<WorkerC
 
             WorkerConnection::connect_alpha_locator_async(
                 &alpha_locator,
-                &ConnectionParameters::new(worker_type, components)
+                &ConnectionParameters::new(worker_type)
                     .using_tcp()
-                    .using_external_ip(true),
+                    .using_external_ip(true)
+                    .generate_component_vtable(),
```
davedissian: Nit: Would prefer naming this to something that isn't "vtable", as that's an implementation detail. Maybe `enable_internal_serialization()`

jamiebrynes7: Yup - seems sensible, will change

PR #80, opened 2019-03-05
Finish Snapshot API

### LINKED ISSUE #13: Flesh out Snapshot APIs
The current implementation is only halfway there. 

Probably depends on the EntitySnapshot

jamiebrynes7:
This PR completes the Snapshot API implementation. Also implemented `EntityQueryResponse` as a driveby. 

Work left to do:
- [x] Move `generate_snapshot` to examples. 
- [x] Find a way to test snapshot output and input programatically. (This requires code generation so will probably require a separate project)
- [x] Some TODOs littered in the codebase. 

Unresolved questions: 

- [x] Currently the implementation pre-supposes the `ComponentDatabase` exists. This should probably be refactored to have the option for custom vtables. Not sure how urgent this is? 

Fixes #13


randomPoison:
Haven't yet had a chance to review the PR, but with regards to your unresolved question:

> Currently the implementation pre-supposes the ComponentDatabase exists. This should probably be refactored to have the option for custom vtables.

Could you elaborate on what you mean here? Do we potentially want a different serialization solution when working with snapshots vs when running a worker?

davedissian:
Agreed with @randomPoison - I would say that it's fine to assume a component database exists. It doesn't even make sense to serialize/deserialize a snapshot without one anyway without providing a custom vtable, but why should vtable logic be different compared to the connection?

jamiebrynes7:
This is more related to if a user wants to write their own code generation we are locking them into `ComponentDatabase` and `inventory`. Until we have a strong use case against this - I'm fine with leaving this in.

davedissian:
Yeah, I'd wait til then. However, I think we may want to allow users to disable the component database and pass in their own `Schema` objects.

jamiebrynes7:
Sounds like a good use case for [Cargo features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section). I think it should be that the component vtable approach is opt-out rather than opt-in. 

Might end up looking something like the C# SDKs structure 😉

spatialos-sdk/src/worker/entity.rs
```diff
@@ -36,36 +45,151 @@ impl Entity {
...
+    pub(crate) fn add_raw(&mut self, component: &Worker_ComponentData) -> Result<(), String> {
+        let id = component.component_id;
+
+        self.pre_add_check(id)?;
+
+        // Call copy on the component data. We don't own this Worker_ComponentData.
+        let vtable = self.database.get_vtable(id).unwrap();
+        let copy_data_func = vtable
+            .component_data_copy
+            .unwrap_or_else(|| panic!("No component_data_free method defined for {}", id));
```
randomPoison: It looks like you need to unwrap here because all of the fields on `Worker_ComponentVtable` are optional (it's a bit unclear to me why the C bindings are generated this way but 🤷). In practice, though does this make sense? Is it ever valid for any of these vtable functions to not be provided? I think the implementation of the `Component` trait currently guarantees that all the components would be supplied. Even beyond that, does the C SDK allow vtable function pointers to be null? I would expect that to be an error.

If it's the case that all fields will always be present, then it would be good to use a different `VTable` type that has the fields as non-optional, statically avoiding the need for this `unwrap`.

spatialos-sdk/src/worker/entity.rs
```diff
@@ -36,36 +45,151 @@ impl Entity {
             user_handle: data_ptr as *mut _,
         };
 
+        self.components.insert(C::ID, raw_data);
+
+        Ok(())
+    }
+
+    pub(crate) fn add_raw(&mut self, component: &Worker_ComponentData) -> Result<(), String> {
```
randomPoison: `add_raw` should probably be marked as `unsafe` unless we want to add more validity checks around how `component` is used. Specifically, we don't currently do anything to check that `component.user_handle` is non-null (or otherwise valid) before trying to use it, and I'm not sure it makes a ton of sense to add those checks. The safety of `add_raw` is entirely dependent on the validity of `component`, so it seems incorrect for it to be a "safe" function.

spatialos-sdk/src/worker/entity.rs
```diff
@@ -36,36 +45,151 @@ impl Entity {
...
+        if self.components.contains_key(&id) {
+            return Err(format!(
+                "Duplicate component with ID {} added to `Entity`.",
+                id
+            ));
+        }
+
+        if !self.database.has_vtable(id) {
+            return Err(format!(
+                "Could not find a vtable implementation for component {}",
```
randomPoison: Does it make sense to return an error if the vtable is missing? With the way the SDK is setup currently, this could only be caused by a misconfiguration by the user, i.e. they didn't register a component. It might be more appropriate to panic in this case.

spatialos-sdk/src/worker/entity.rs
```diff
@@ -36,36 +45,151 @@ impl Entity {
...
     }
+}
 
-    pub(crate) fn raw_component_data(&self) -> Vec<Worker_ComponentData> {
-        self.components.values().map(|data| data.raw_data).collect()
+impl Default for Entity {
+    fn default() -> Self {
+        Entity {
+            components: HashMap::new(),
+            database: get_component_database(),
```
randomPoison: You could impl `Default` for `ComponentDatabase` and then you'd be able to derive `Default` for `Entity` again.

spatialos-sdk/src/worker/entity.rs
```diff
@@ -36,36 +45,151 @@ impl Entity {
...
+    pub components: Vec<Worker_ComponentData>,
+    database: ComponentDatabase,
+}
+
+impl RawEntity {
+    pub fn new<'a, I>(original_data: I) -> Self
+    where
+        I: Iterator<Item = &'a Worker_ComponentData>,
+    {
+        let database = get_component_database();;
```
randomPoison: Double `;`? I'm surprised, I thought that was a compile error 🤔

spatialos-sdk/src/worker/snapshot.rs
```diff
@@ -1,43 +1,116 @@
...
 }
 
 impl SnapshotOutputStream {
-    pub fn new<P: AsRef<Path>>(filename: P, params: &SnapshotParameters) -> Self {
+    pub fn new<P: AsRef<Path>>(filename: P) -> Result<Self, String> {
         let filename_cstr = CString::new(filename.as_ref().to_str().unwrap()).unwrap();
 
-        let ptr = unsafe {
-            Worker_SnapshotOutputStream_Create(filename_cstr.as_ptr(), &params.to_worker_sdk())
+        let database = get_component_database();;
```
randomPoison: Nother double `;;`.

test-suite/src/snapshot_integration_tests.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::env;
+use std::f64::EPSILON;
+
+use crate::generated::improbable::*;
+use std::collections::BTreeMap;
+
+#[test]
+pub fn create_and_read_snapshot() {
+    let mut snapshot_path = env::temp_dir();
+    snapshot_path.push("test.snapshot");
```
randomPoison: This can be expressed more succinctly as `let snapshot_path = env::temp_dir().join("test.snapshot");`.

test-suite/src/snapshot_integration_tests.rs
```diff
@@ -0,0 +1,72 @@
...
+        assert!(snapshot.has_next());
+
+        let entity = snapshot.read_entity().expect("Error");
+
+        let position = entity.get::<Position>();
+        assert!(position.is_some());
+        let coords = &position.unwrap().coords;
+        assert!((10.0 - coords.x).abs() < EPSILON);
+        assert!((-10.0 - coords.y).abs() < EPSILON);
+        assert!((0.0 - coords.z).abs() < EPSILON);
```
randomPoison: There's a nifty crate for doing approximately-equals comparisons like this: https://docs.rs/approx/0.3.1/approx/

spatialos-sdk/src/worker/component.rs
```diff
@@ -232,24 +235,72 @@ pub(crate) mod internal {
 
 inventory::collect!(VTable);
 
-// A data structure which represents all known component types. Used to generate an array of vtables to pass
-// to the connection object.
-#[derive(Default)]
+// This is pattern used in the Rust std lib for a singleton object. Sadly we cannot use lazy_static
+// due to the unsafe types within the component vtables. This operates similarly to the stdin()
+// method. We need a singleton instance as inventory appears to only allow iteration once (at
+// least that was the observed behaviour).
```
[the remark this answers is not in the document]
jamiebrynes7: I initially had the multiple iteration solution - but when testing noticed that the second/third/etc object instantiated with a `ComponentDatabase`, the database was empty. 

See https://github.com/jamiebrynes7/spatialos-sdk-rs/blob/cbbc9cb845f387e68cf8b0901e3cbabc712b1c15/spatialos-sdk/src/worker/component.rs#L244

for the implementation at that time. Its possible that I misdiagnosed the issue

spatialos-sdk/examples/generate_snapshot.rs
```diff
@@ -12,7 +9,7 @@ fn main() {
 
     if args.len() != 2 {
         panic!(
-            "Incorrect usage. Expected usage: cargo run --bin generate_snapshot <path-to-snapshot>"
+            "Incorrect usage. Expected usage: cargo run --example generate_snapshot <path-to-snapshot>"
```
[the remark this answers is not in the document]
jamiebrynes7: Probably worth doing as a driveby

spatialos-sdk/src/worker/entity.rs
jamiebrynes7: The `Option<func>` syntax is generated by `bindgen` and my guess is that it signifies that you can pass a `null` ptr down to the C API, which is valid behaviour.

For example, if you want schema objects to be passed back up instead of letting the C API do the deserialization, you give it a vtable full of null-ptrs as the default.

In our implementation, this is never the case - but the underlying C object is like that for a reason. That doesn't mean we can't force `VTable` instances to fully declare it, particularly if we plan to ring fence the default generated code and all associated with it behind a feature flag.

spatialos-sdk/src/worker/entity.rs
jamiebrynes7: Yup, nice catch

spatialos-sdk/src/worker/entity.rs
jamiebrynes7: Yeah a panic may be more appropriate - also reduces the need to `unwrap()` empty returns

spatialos-sdk/src/worker/entity.rs
jamiebrynes7: I guess its interpreted as `get_component_database()` and then a no-op?

test-suite/src/snapshot_integration_tests.rs
jamiebrynes7: Oh nice! :smile:

spatialos-sdk/src/worker/component.rs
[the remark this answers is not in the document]
jamiebrynes7: Well I wasn't able to repro this again 🤔 

I did write most of this quite late at night which may explain it! Will be able to simplify this greater

spatialos-sdk/src/worker/entity.rs
randomPoison: Does the C API support partially-null vtables, though? It seems like vtables either have to be fully specified or fully `null`. Is it ever meaningful/valid for only some of the function pointers in the vtable to be null?

jamiebrynes7: I'm not sure. cc @davedissian ?

davedissian: Yes, it does fully support it (although in general, not sure what the use cases would be). Perhaps you want to have specialized serialization functions for components, but commands go via a general reflection based serializer (?)

spatialos-sdk/src/worker/entity.rs
```diff
@@ -153,7 +159,7 @@ impl RawEntity {
 
         RawEntity {
             components: new_data,
-            database,
+            database: &component::DATABASE,
```
randomPoison: Now that `ComponentDatabase` is a singleton (again) it shouldn't be necessary to store references to it in a struct.

spatialos-sdk/src/worker/parameters.rs
```diff
@@ -71,7 +74,7 @@ impl ConnectionParameters {
     }
 
     pub fn enable_internal_serialization(mut self) -> Self {
-        self.components = Some(Default::default());
+        self.components = Some(&DATABASE);
```
randomPoison: Also not necessary to store the database here, I think.

spatialos-sdk/src/worker/entity.rs
```diff
@@ -5,26 +5,25 @@ use std::collections::HashMap;
 use std::ptr;
 use std::slice;
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct Entity {
     components: HashMap<ComponentId, Worker_ComponentData>,
-    database: ComponentDatabase,
+    database: &'static ComponentDatabase,
```
randomPoison: Also shouldn't be necessary to store a reference to the database here.

spatialos-sdk/src/worker/entity.rs
randomPoison: Interesting! In that case, it makes sense to keep the C type definitions the same and make an internal vtable type where the fields aren't optional to better represent that all of the functions will always be present for the Rust types.

PR #85, opened 2019-03-21
Add Entity Builder

### LINKED ISSUE #25: Add some sort of Entity Builder API
It's very idiomatic to use the builder pattern in Rust and this pattern has worked well for defining entity templates in the past/

jamiebrynes7:
This PR introduces an `EntityBuilder` which helps users create entities and cuts down the verbosity required to do so.

It includes: 
- Forcing a user to provide a position (required component).
- Automatic building of the `EntityACL` component (required component).
- Making users specify a write-access layer when adding a component.
- Providing special methods for other well-known components (Metadata and Persistence).

These four features makes it much less error-prone when creating entities. However, there is still one place to trip up on:

- Setting read layers is still done in a separate method. This is easy to mess up (by omission). We could make this part of the EntityBuilder constructor to force users to pass this in, but the constructor will start to get very long.

Note that there is a little bit of hackery in the implementation as we need to "access" code generated types (Position, EntityAcl, Metadata, EntityAcl) but do not have the generated code available at compile time. To get around this, we manually serialize the data (since its a well known shape) then deserialize inside the `Entity` itself. (At runtime we do expect to have these types defined). Do we expect the `add_serialized` method on `Entity` to be used elsewhere? Or can it be folded into the `EntityBuilder` as an implementation detail? 

Fixes #25

### DIFF
--- a/spatialos-sdk/src/worker/entity_builder.rs
+++ b/spatialos-sdk/src/worker/entity_builder.rs
@@ -30,7 +30,9 @@ impl EntityBuilder {
             error: None,
         };
 
-        builder.write_permissions.insert(POSITION_COMPONENT_ID, position_write_layer.into());
+        builder
+            .write_permissions
+            .insert(POSITION_COMPONENT_ID, position_write_layer.into());
 
         builder
     }
@@ -69,6 +71,15 @@ impl EntityBuilder {
         Ok(self.entity)
     }
 
+    // A workaround for not having access to generated code types here. The shape of Position
+    // & EntityAcl are well known, so we can manually serialize them and pass them into the
+    // Entity in SchemaComponentData form.
+    //
+    // This does then expect that there is a valid deserialize method defined for both components
+    // in the vtable.
+    //
+    // If this invariant is broken, then the EntityBuilder is broken. Should we assert against this
+    // before we call `entity.add_serialized`?
     fn serialize_position(&self) -> SchemaComponentData {
         let mut position_schema = SchemaComponentData::new(POSITION_COMPONENT_ID);
         let position_fields = position_schema.fields_mut();
--- a/spatialos-sdk/src/worker/entity_builder.rs
+++ b/spatialos-sdk/src/worker/entity_builder.rs
@@ -24,7 +24,7 @@ pub struct EntityBuilder {
     write_permissions: HashMap<ComponentId, String>,
     read_permissions: Vec<String>,
 
-    error: Option<String>
+    error: Option<String>,
 }
 
 impl EntityBuilder {
@@ -56,7 +56,11 @@ impl EntityBuilder {
         self.set_write_access(PERSISTENCE_COMPONENT_ID, write_layer)
     }
 
-    pub fn set_metadata<T: Into<String>, U: Into<String>>(mut self, entity_type: T, write_layer: U) -> Self {
+    pub fn set_metadata<T: Into<String>, U: Into<String>>(
+        mut self,
+        entity_type: T,
+        write_layer: U,
+    ) -> Self {
         self.metadata = Some(entity_type.into());
         self.set_write_access(METADATA_COMPONENT_ID, write_layer)
     }
@@ -144,7 +148,9 @@ impl EntityBuilder {
     fn serialize_metadata(&self) -> SchemaComponentData {
         let mut metadata_schema = SchemaComponentData::new(METADATA_COMPONENT_ID);
         let metadata_fields = metadata_schema.fields_mut();
-        metadata_fields.field::<SchemaString>(1).add(self.metadata.as_ref().unwrap());
+        metadata_fields
+            .field::<SchemaString>(1)
+            .add(self.metadata.as_ref().unwrap());
 
         metadata_schema
     }
--- a/spatialos-sdk/src/worker/entity_builder.rs
+++ b/spatialos-sdk/src/worker/entity_builder.rs
@@ -9,32 +9,37 @@ use crate::worker::{
 };
 use std::collections::HashMap;
 
-const POSITION_COMPONENT_ID: ComponentId = 54;
 const ENTITY_ACL_COMPONENT_ID: ComponentId = 50;
+const METADATA_COMPONENT_ID: ComponentId = 53;
+const POSITION_COMPONENT_ID: ComponentId = 54;
+const PERSISTENCE_COMPONENT_ID: ComponentId = 55;
 
 pub struct EntityBuilder {
-    position: (f64, f64, f64),
     entity: Entity,
+
+    position: (f64, f64, f64),
+    is_persistent: bool,
+    metadata: Option<String>,
+
     write_permissions: HashMap<ComponentId, String>,
     read_permissions: Vec<String>,
-    error: Option<String>,
+
+    error: Option<String>
 }
 
 impl EntityBuilder {
     pub fn new<T: Into<String>>(x: f64, y: f64, z: f64, position_write_layer: T) -> Self {
-        let mut builder = EntityBuilder {
-            position: (x, y, z),
+        let builder = EntityBuilder {
             entity: Entity::new(),
+            is_persistent: false,
+            metadata: None,
+            position: (x, y, z),
             write_permissions: HashMap::new(),
             read_permissions: Vec::new(),
             error: None,
         };
 
-        builder
-            .write_permissions
-            .insert(POSITION_COMPONENT_ID, position_write_layer.into());
-
-        builder
+        builder.set_write_access(POSITION_COMPONENT_ID, position_write_layer)
     }
 
     pub fn add_component<C: Component, T: Into<String>>(mut self, data: C, write_layer: T) -> Self {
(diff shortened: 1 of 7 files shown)


jamiebrynes7:
An example of the EntityBuilder API: 

```rust
    let entity = EntityBuilder::new(0.0, 0.0, 0.0, "rusty")
        .add_component(
            example::Rotate {
                angle: rng.gen_range(0.0, 2.0 * f64::consts::PI),
                radius: rng.gen_range(20.0, 100.0),
                center_x: rng.gen_range(-50.0, 50.0),
                center_y: 0.0,
                center_z: rng.gen_range(-50.0, 50.0),
            },
            "rusty",
        )
        .add_component(
            improbable::Metadata {
                entity_type: "Rotator".to_owned(),
            },
            "rusty",
        )
        .set_read_access(&["rusty"])
        .set_write_access(EntityAcl::ID, "rusty")
        .build()
        .unwrap();
```

randomPoison:
Ah heck, looks like you're running into exactly the sort of issue #56 is intended to fix! Unfortunately, my work on that isn't nearly far enough along for it to be worth waiting on, so I think moving forward with the hacky stuff here makes sense. Once #56 is closed we'll be able to remove `add_serialized` entirely.

randomPoison:
Is there a meaningful difference between `Entity` and `EntityBuilder`? `Entity` already has the `add` method for attaching components (which `EntityBuilder` is using internally), so it seems like `Entity` builder just changes/extends the API of `Entity`, in which case would it make more sense to change `Entity` directly to have the improved API?

jamiebrynes7:
So there's one thing that would stop me from merging the two APIs - when would we serialize the `EntityAcl`? 

You could have a `build_acl()` method, but nothing stops a user adding more components afterward and shooting themselves in the foot. Its an instance where not only do you have to remember to call a specific method, it has to be at a specific time.

(On that topic, we may want to remove the public facing `add_component` method on `Entity` to prevent similar errors here) 

In theory #56 solves that issue since you never have to serialize, but I'm not 100% convinced that we should have a crate contain the generated code and user code which operates on that generated code (a discussion for another day).

randomPoison:
Ah, building the `EntityAcl` is a good point! In that case, I agree that it would make more sense to remove the public `Entity::add` method (or at least make it private to the crate) and make it so that `EntityBuilder` is the only way to construct the serialized representation of an entity.

That would also address a concern I'd been having about `Entity`, that it's current purpose is a bit unclear. It's both used to create the serialized representation of an entity, but it's also used to read serialized entity data sent from the runtime (which can't be modified). But if `Entity` represents the immutable, serialized version of data, with `EntityBuilder` creating the data in the worker or already-serialized data coming in from the runtime, I think that makes for a pretty clean API :+1:

spatialos-sdk/src/worker/entity_builder.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub fn set_metadata<T: Into<String>, U: Into<String>>(
+        mut self,
+        entity_type: T,
+        write_layer: U,
+    ) -> Self {
+        self.metadata = Some(entity_type.into());
+        self.set_write_access(METADATA_COMPONENT_ID, write_layer)
+    }
+
+    pub fn set_read_access<T: AsRef<str>>(mut self, layers: &[T]) -> Self {
```
randomPoison: I'm not sure about the way read/write access are being specified :thinking: I like that you've set the API to reflect the new layers system (e.g. it only allows a single write layer to be specified per component), but I don't know that separating the read layers from specifying the component is a good idea.

It seems like the ideal approach would be to specify the component, its write layer, and its read layer(s) all at the same time. The issue is how to do that in an ergonomic way. There are a couple of approaches I can think of:

We could have a separate builder for for components, so it would look something like:

```rust
let entity = EntityBuilder::new()
    .add(
        ComponentBuilder::new(my_component)
            .write_layer("foo")
            .read_layers(&["foo", "bar", "baz"]),
    );
```

The most obvious drawback here is that it's really verbose and annoying to use. A potentially less verbose alternative (that could be built on top of the component builder) is to use a macro:

```rust
let entity = EntityBuilder::new()
    .add(component!(
        my_component,
        write: "foo",
        read: ["foo", "bar", "baz"],
    ));
```

We could possibly extend this even further by providing a macro for the entire entity builder:

```rust
let entity = entity![
    { my_component, write: "foo", read: ["foo", "bar", "baz"] },
    { other_component, write: "bar", read: ["bar", "baz"] },
];
```

I generally avoid using macros, but in this case I think it greatly improves the ergonomics, and can just use the `EntityBuilder`/`ComponentBuilder` API under the hood.

spatialos-sdk/src/worker/entity_builder.rs
```diff
@@ -0,0 +1,161 @@
...
+        builder.set_write_access(POSITION_COMPONENT_ID, position_write_layer)
+    }
+
+    pub fn add_component<C: Component, T: Into<String>>(mut self, data: C, write_layer: T) -> Self {
+        if let Err(e) = self.entity.add(data) {
+            self.error = Some(e);
+        };
+
+        self.write_permissions.insert(C::ID, write_layer.into());
+        self
```
randomPoison: I'm going to go out on a limb and make a bit of a wild request, but could we *not* return `self` from the methods on `EntityBuilder`? I actually just published an article about why this is a bad pattern: https://randompoison.github.io/posts/returning-self. I'd like us to move away from building method chaining into our API, since there are a number of places where we are doing so already.

spatialos-sdk/src/worker/entity_builder.rs
jamiebrynes7: Read access is declared at an entity level rather than a component level :wink:

randomPoison: Oh right, sorry 🤦‍♀️ Well in that case, maybe make the write layer parameter an `Option<String>`? Or even better an `Option<T> where T: Into<String>` so you can do `Some("foo")`.

I also wonder if `set_write_access` is needed at all? Since you set the write access of a component type when it's added, do we need an additional way to set write access?

randomPoison:
Would it make sense to automatically build `read_permissions` from layers that are given write access to the components? i.e. if you don't call `set_read_access`, it defaults to given read access to any layers that are given write access to one of the component types? I imagine it's pretty rare to give a layer write access to a component on an entity without giving that layer read access to the entity.

Actually, is that *ever* a valid thing to do? Should we have `EntityBuilder` **always** give read access to any layer that has write access to one of the components? We could then change `set_read_access` to something like `add_read_access` to allow users to grant additional layers read access without allowing them to forget to give read access to layers that have write access. I haven't used SpatialOS enough to know if this makes sense, though. Thoughts?

spatialos-sdk/src/worker/entity_builder.rs
jamiebrynes7: Hmm - what would the use case be for declaring that nothing can ever write to a component? 

In practice, this can be achieved by passing in a `""` argument - but I'd prefer not to encode it as part of the API. 

---
The reason I've added `set_write_access` is mainly for letting a user add an owner for the `EntityAcl` component, this could be more clearly focused to: 
`set_entity_acl_write_access`?

randomPoison: It would probably be better to change the generic bound here to `T: Into<String>`, since the only thing we ever do with the layer names here is convert them into `String`s.

jamiebrynes7:
I think you are correct, write access should imply read access - there is one edge case I can think of, but I'm not sure what the load balancer would do in that case: 

If you have an entity which you delegate read access to a specific worker (i.e. - `workerId: RustWorker-1`) but then delegate write access to its entire layer (`rusty`). I don't know if the load balancer can resolve that. Obviously, if you were doing this, you'd just delegate write access to that specific worker as well - but the previous configuration could also be valid..?

> We could then change set_read_access to something like add_read_access to allow users to grant additional layers read access without allowing them to forget to give read access to layers that have write access.

This does seem sensible though

spatialos-sdk/src/worker/entity_builder.rs
randomPoison: I can imagine cases where you'd want to create read-only components in order to associate immutable data with an entity. I expect that some of the built-in components are often never modified (e.g. `Persistence` and `Metadata`). Though I get your point that it's probably waaaaaaaaaaaay more common to assign write access to a component, so maybe have a separate `add_immutable` method that doesn't take a write parameter? That would still allow users to add immutable components while making it more clear that they're doing so (instead of accidentally forgetting to assign a write layer).

randomPoison: Yeah, I think something like `set_acl_write_access` would be clearer (and simpler, since it wouldn't need the component ID parameter) 👍

randomPoison:
For the edge case you're imagining, is that a valid use case or indicative of a configuration error on the part of the user?

Also, I can vaguely imagine a case where it makes sense to grant write-only access to a component (maybe for the purpose of security and information hiding), but I think it makes sense to disallow doing so until we have a concrete use case, since it seems too easy to do by mistake and end up creating an invalid configuration.

jamiebrynes7:
@randomPoison So I've made the following changes:

- Made `add_component` internal to the crate on `Entity`.
- Removed the builder method chaining from the `EntityBuilder`.
- Changed `set_write_access` to `set_entity_acl_write_access`.
- Whenever a layer is added as write access, it is added as read as well.
- Changed `set_read_acess` to `add_read_access`

PR #87, opened 2019-03-21
Make ReserveEntityIdsResponse contain an iterator

### LINKED ISSUE #65: ReserveEntityIdsResponse should return the list of entities as an iterator
Currently (as of my PR which implements send_component_update), `ReserveEntityIdsResponse` defines a range of entity IDs as: "first entity ID" and "number of entity IDs".

We can probably turn this into some kind of generator pattern, that allows someone to `.take()` an entity ID from the reserved range.

jamiebrynes7:
Reworked the `ReservedEntityIdsRange` to be an iterator and hid the internals. 

Fixes #65

### DIFF
--- a/spatialos-sdk/src/worker/op.rs
+++ b/spatialos-sdk/src/worker/op.rs
@@ -647,3 +647,25 @@ impl<'a> CommandResponse<'a> {
         &self.response.schema_type
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::ReservedEntityIdRange;
+
+    #[test]
+    fn reserved_entity_id_range_iterator_contains_correct_count() {
+        let range = ReservedEntityIdRange::new(10, 54);
+        assert_eq!(54, range.count());
+    }
+
+    #[test]
+    fn reserved_entity_id_range_iterator_returns_sequential_ids() {
+        let mut current_id :i64 = 1;
+        let range = ReservedEntityIdRange::new(current_id, 10);
+
+        for entity_id in range {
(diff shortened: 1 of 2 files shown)


spatialos-sdk/src/worker/op.rs
```diff
@@ -482,8 +479,35 @@ pub struct ReserveEntityIdsResponseOp {
...
 #[derive(Debug)]
 pub struct ReservedEntityIdRange {
-    pub first_entity_id: EntityId,
-    pub number_of_entity_ids: u32,
+    current: i64,
+    consumed: u32,
+    reserved: u32,
+}
+
+impl ReservedEntityIdRange {
```
davedissian: Why not instead use `std::ops::Range` and delete the custom type entirely?

jamiebrynes7: Because I didn't know that existed 😂

jamiebrynes7: It seems very appropriate - will refactor

jamiebrynes7: Sadly we lose the `Iterator` impl as `EntityId` must implement [`Step`](https://doc.rust-lang.org/std/iter/trait.Step.html) to get the blanket implementation for `Range<EntityId>`.

Unfortunately, `Step` is nightly and I don't want to force us onto a nightly version

jamiebrynes7: What I can do - is add a TODO to replace with `Range<EntityId>` once `Step` is stabilised.

PR #88, opened 2019-03-22
Fix Load Metric

jamiebrynes7:
Fixed a bug where load was reported as 0 always due to taking a reference to a local variable.


jamiebrynes7:
I reckon that proper testing should catch these sort of bugs. That is - convert to the C repr and then assert that the data inside is what is expected.

What sort of layer of encapsulation were you thinking of?

PR #89, opened 2019-04-04
WIP: Serialization rewrite

randomPoison:
This is a chunk of work that I'm trying to split off from my work on #56. In order to simplify the code that we need to generate for serialization, I'm redoing the serialization traits. The work here isn't done yet, but I wanted to start getting some feedback, both in terms of the implementation and the process we want around merging these changes.

In particular, I'd like to be able to merge these changes ahead of the rest of the changes for #56 in order to reduce the size of an already massive PR. However, I don't want to update the existing code generation to use the new serialization traits, since that would be difficult to do (at least for me) and would be wasted work since I'm already rewriting the code generation logic. Does anyone object to breaking the SDK on master for a bit while I finish the rest of the rewrite work? If that's going to be a problem, is there a better way to break up the rewrite such that I don't have to do everything in one giant PR?

I wanted to get this PR up early so that we can decide on how to handle this before I'm actually ready to merge anything.

---

The other reason for getting this PR up early is so that I can give further feedback to @davedissian about the C SDK serialization, as well as get feedback from him about whether or not I'm encoding the schemalang semantics correctly.

So far I've only put the core logic in place for deserializing types from `SchemaObjects`, but the design should work fine for serialization. I've broken serialization into the following traits:

* `SchemaType` which provides the `from_field` method, allowing a type to be deserialized from a single field on a `SchemaObject`.
* `SchemaIndexType` which provides the `field_count` and `index_field` methods.
* `SchemaArrayType` which provides the `from_field_array` method for types that can be deserialized directly from an array field (i.e. primitive types).

Additionally, I've defined a `SchemaObjectType` to represent objects that can be serialized directly from an entire `SchemaObject` (i.e. user-defined schema `type`s). `SchemaType` and `SchemaIndexType` are implemented automatically for all types that implement `SchemaObjectType`. Code generation for user-defined types will only generate an impl for `SchemaObjectType`.

Splitting up the traits like this allows us to statically encode the semantics of schemalang types in the corresponding Rust types. Specifically:

* Collection types (`Option<T>`, `Vec<T>`, and `BTreeMap<K, V>`) only implement `SchemaType`, but require that their type parameters implement `SchemaIndexType`, enforcing that you can't create nested collection types (e.g. `Vec<Option<T>>` can't be serialized because it's not a valid schema type).
* Types that implement `SchemaObjectType` implement `SchemaIndexType` but not `SchemaArrayType`, since the C API provides `Schema_GetObject`, `Schema_GetObjectCount`, and `Schema_IndexObject`, but not `Schema_GetObjectList` (reasonably so, as that wouldn't make a ton of sense).

---

In terms of code generation, the advantage of all of this is that the code we need to generate becomes some pretty uniform boilerplate. For example, given this schema definition:

```
type MyType {
    int32 foo = 0;
    string bar = 1;
    AnotherType baz = 2;
    vec<CoolType> quux = 3;
}
```

The deserialization code would be:

```rust
pub struct MyType {
    pub foo: i32,
    pub bar: String,
    pub baz: AnotherType,
    pub quux: Vec<CoolType>,
}

impl SchemaObject for MyType {
    fn from_schema_object(schema_object: &SchemaObject) -> Self {
        Self {
            foo: schema_object.field::<SchemaInt32>(0),
            bar: schema_object.field::<String>(1),
            baz: schema_object.field::<AnotherType>(2),
            quux: schema_object.field::<Vec<CoolType>>(3),
        }
    }
}
```

Every field is deserialized the same way, deferring to the `SchemaType` implementation for field. Since all types must implement `SchemaType`, we don't need to know anything about the type of the field (other than the name of the type) in order to generate the serialization code.

---

There's still a lot that I need to do to finish up this work (I haven't actually added `SchemaArrayType` yet, for example), but I'm happy to take any feedback!


jamiebrynes7:
> In particular, I'd like to be able to merge these changes ahead of the rest of the changes for #56 in order to reduce the size of an already massive PR. However, I don't want to update the existing code generation to use the new serialization traits, since that would be difficult to do (at least for me) and would be wasted work since I'm already rewriting the code generation logic. Does anyone object to breaking the SDK on master for a bit while I finish the rest of the rewrite work? If that's going to be a problem, is there a better way to break up the rewrite such that I don't have to do everything in one giant PR?

On this, if we branch off of master, into a `experimental/codegen-rewrite` or `feature/codegen-rewrite` then we can point PRs into that branch and review things incrementally. Then when its all good we can merge the experimental branch into master. There isn't a huge volume of PRs right now so we shouldn't risk too much desync (nothing that a good `git rebase` won't solve!).

I'd prefer not to have master be broken

randomPoison:
Sounds good! Then once this work is ready and approved, I'll convert it into `experimental/codegen-rewrite` (or something like that) and future PRs will be made against that branch :+1: I'll leave the current branch and PR as-is, but won't merge it into master once it's approved.

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+pub trait SchemaIndexType: SchemaType {
+    fn field_count(object: &SchemaObject, field: FieldId) -> u32;
+
+    fn index_field(object: &SchemaObject, field: FieldId, index: u32) -> Self::RustType;
+}
+
+pub trait SchemaListType: SchemaIndexType {
+    fn get_field_list(object: &SchemaObject, field: FieldId, data: &mut Vec<Self::RustType>);
```
jamiebrynes7: Why do we expect the caller to pass in a `Vec` to mutate rather than returning the `Vec`? 

Its slightly asymmetric compared to `from_field` and `index_field`

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+// =================================================================================================
+// Schema Conversion Traits
+// =================================================================================================
+
+pub trait SchemaType: Sized {
+    type RustType: Sized;
+
+    fn from_field(object: &SchemaObject, field: FieldId) -> Self::RustType;
```
jamiebrynes7: Nit: `get_field` to be symmetric with `get_field_list`

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+}
+
+pub trait SchemaIndexType: SchemaType {
+    fn field_count(object: &SchemaObject, field: FieldId) -> u32;
```
jamiebrynes7: `get_field_count` ?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+        unsafe { Schema_GetBoolCount(object.internal, field) }
+    }
+
+    fn index_field(object: &SchemaObject, field: FieldId, index: u32) -> Self::RustType {
+        let raw = unsafe { Schema_IndexBool(object.internal, field, index) };
+        raw != 0
+    }
+}
+
+// TODO: This is only valid if schema bools are guaranteed to be 0 or 1. If schema
```
jamiebrynes7: I assume this is referring to the types defined above? Not the `Option<T>`? 

And I believe this is the case. If we receive a value outside of `{ 0, 1 }` it should probably cause a panic

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+
+impl<K, V> SchemaType for BTreeMap<K, V>
+where
+    K: SchemaIndexType,
+    V: SchemaIndexType,
+    K::RustType: Ord,
+{
+    type RustType = BTreeMap<K::RustType, V::RustType>;
+
+    fn from_field(object: &SchemaObject, field: FieldId) -> Self::RustType {
```
jamiebrynes7: I think this is incorrect. I believe the map is represented in schema-land as N key-pairs at the field index of the map. So you have something like: 

```
SchemaObject
	- keyPair1
		- Key
		- Value
	-keyPair2
		-Key
		- Value
```

I think the correct code should be: 

```rust
fn from_field(object: &SchemaObject, field: FieldId) -> Self::RustType {
	let count = unsafe { Schema_GetObjectCount(object.internal, field) };

	let mut result = BTreeMap::new();
	for index in 0..count {
		let kv_pair = &SchemaObject { internal : unsafe { Schema_IndexObject(object.internal, field, index) } };
		let key = K::from_field(kv_pair, SCHEMA_MAP_KEY_FIELD_ID);
		let value = V::from_field(kv_pair, SCHEMA_MAP_VALUE_FIELD_ID);
        result.insert(key, value);
	}

    result
}
```

jamiebrynes7: For reference, here's how we do it in C# - https://github.com/spatialos/gdk-for-unity/blob/master/test-project/Assets/Generated/Source/improbable/gdk/tests/ExhaustiveMapKey.cs#L1414

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+    fn from_field(object: &SchemaObject, field: FieldId) -> Self::RustType;
+}
+
+pub trait SchemaIndexType: SchemaType {
+    fn field_count(object: &SchemaObject, field: FieldId) -> u32;
+
+    fn index_field(object: &SchemaObject, field: FieldId, index: u32) -> Self::RustType;
+}
+
+pub trait SchemaListType: SchemaIndexType {
```
davedissian: Why must a `SchemaListType` implement a `SchemaIndexType`? Also, isn't this supposed to be `SchemaArrayType` according to your description?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+    SchemaEnum,
+    Schema_GetEnum,
+    Schema_IndexEnum,
+    Schema_GetEnumCount,
+    Schema_AddEnum,
+    Schema_AddEnumList,
+    Schema_GetEnumList,
+);
+
+impl SchemaType for EntityId {
```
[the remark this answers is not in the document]
randomPoison: So we probably *could* use the macro here, if we added an `Into` conversion between `EntityId` and `i64`. I was hesitant to do that, though, because I'd like to make `EntityId` more strongly-typed (i.e. disallow converting between `i64` and `EntityId`, effectively treat `EntityId` as it's own type), but I don't know if that's a reasonable thing to do. Thoughts?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
+use crate::worker::{component::ComponentId, EntityId};
+use spatialos_sdk_sys::worker::*;
+use std::collections::BTreeMap;
+
+pub type FieldId = u32;
+
+#[derive(Debug)]
+pub struct SchemaComponentUpdate {
+    pub component_id: ComponentId,
+    pub internal: *mut Schema_ComponentUpdate,
```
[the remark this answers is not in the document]
randomPoison: No, this should be private. I'm going to do a pass in a bit to improve encapsulation and see if I can make `SchemaObject` actually safe since it currently has some soundness holes.

spatialos-sdk/src/worker/schema.rs
randomPoison: I'm looking ahead to applying component updates. If you have a component with a `Vec` field and an update for that component, this API should allow us to update the existing `Vec` field without needing an additional allocation. That said, maybe it'd be useful to have an additional method that just returns a new `Vec`, since that's a common use-case anyway 🤔

spatialos-sdk/src/worker/schema.rs
randomPoison: It might actually be good to change the name of `get_field_list` instead 🤔

spatialos-sdk/src/worker/schema.rs
randomPoison: First off: Woops, I left that comment on the wrong impl block. That comment is meant to be applied to `impl SchemaListType for bool`, which somehow got lost 😓 

As for why that would be unsound: The Rust `bool` type is represented as an 8 bit unsigned integer with the value 0 or 1. As such, it's [undefined behavior](https://doc.rust-lang.org/beta/reference/behavior-considered-undefined.html) for a `bool` to have any other value. So if a bool in a schema object can have a value other than 0 or 1, we have can't directly transmute it to a bool, we have to manually convert it to `bool` (which is what the `SchemaType` and `SchemaIndexType` impls for `bool` currently do). The problem with implementing `SchemaListType` for `bool` is that we don't have a chance to do that conversion, we're directly populating a `Vec<bool>` from the schema object's data, so we need to guarantee that doing so will only produce valid values. If the schema API semantics don't guarantee that bools will only ever be 0 or 1, then we can't implement `SchemaListType` for `bool`.

spatialos-sdk/src/worker/schema.rs
randomPoison: Ah heck, I completely misunderstood what the docs were saying 🙃 Thanks for catching this!

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct SchemaObject {
+    internal: *mut Schema_Object,
+}
+
+impl SchemaObject {
+    pub fn field<T: SchemaType>(&self, field: FieldId) -> T::RustType {
```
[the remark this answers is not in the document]
randomPoison: That's reasonable! I actually think the traits need some renaming to match Rust's style guide. Currently they're all nouns, but I think the convention is to have traits be verbs? I'll have to confirm that, though. Either way, `SchemaFieldType` is a more accurate name for sure 👍

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,582 @@
...
+// and should be removed.
+impl<T: SchemaIndexType> SchemaType for Option<T> {
+    type RustType = Option<T::RustType>;
+
+    fn from_field(object: &SchemaObject, field: FieldId) -> Self::RustType {
+        let count = T::field_count(object, field);
+        match count {
+            0 => None,
+            1 => Some(T::from_field(object, field)),
+            _ => panic!(
```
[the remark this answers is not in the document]
randomPoison: Ah, that's interesting! I had assumed that the runtime would reject malformed component data, but I guess not? If that's not the case, at least, then I agree that we shouldn't panic in this case. This would be a good thing to clarify in the serialization docs, though. I'll add a question about it to the forum thread as a reminder 😁

randomPoison:
I've renamed the traits to `SchemaField`, `IndexedField`, and `ArrayField`. I'm still not 100% set on the names, but hopefully these better captures what each of the traits represent while being reasonably succinct. Let me know if you have other suggestions, though!

randomPoison:
I've sealed `SchemaField`, `IndexedField`, and `ArrayField` (following the [future proofing guide](https://rust-lang-nursery.github.io/api-guidelines/future-proofing.html)). The idea is that those three traits are implemented only for specific primitive and collection types, and external crates should only implement `SchemaObjectType` for their custom schema types (well, there will also be a similar `SchemaEnum` trait, but same idea). Sealing the schema traits ensures that external code can't create new, erroneous implementations for new types.

randomPoison:
So I think I've figured out how to make `SchemaObject` sound. I've added a lifetime parameter, such that it's now `SchemaObject<'a>`. Depending on where the schema object is coming from, this lifetime reflects one of two things:

* When dealing with schema data coming from the runtime from a worker op (e.g. new component data, component updates, etc.), the lifetime reflects the parent object that owns the schema data (most likely the worker op), and ensures that the schema object does not outlive the data it is borrowing.
* When the worker is creating the schema object (e.g. when sending a component update), the lifetime reflects the lifetime of data it has borrowed from the corresponding Rust type.

This latter point is the interesting one, as it's how we're able to safely use the `Schema_Add*List` functions. Since the `Schema_Add*List` functions don't copy any data, we need to statically ensure that the borrowed data outlives the schema object. Fortunately for us, Rust's ownership system allows us to do exactly that. I've updated the relevant traits and methods to correctly associate the lifetime parameter on `SchemaObject` with the data it's borrowing, and have confirmed that trying to populate a schema object with data that doesn't live long enough generates a compiler error. I'm also going to try to setup [compiletest](https://github.com/laumann/compiletest-rs) so that we can have some tests continuing to verify that this is setup correctly.

spatialos-sdk/src/worker/schema.rs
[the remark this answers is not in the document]
jamiebrynes7: I don't think its _that_ crazy to provide the `Into` conversion. The C lib just has a type alias.

spatialos-sdk/src/worker/schema.rs
jamiebrynes7: I wonder if pooling the vecs would be a better approach in this case if we want to maintain a symmetric API. 

In any case - this is a optimisation that we aren't sure if we need to make just yet :wink:

spatialos-sdk/tests/compiletest.rs
```diff
@@ -0,0 +1,14 @@
+use compiletest_rs::common::{Config, Mode};
+use std::path::PathBuf;
+
+#[test]
+fn compile_fail() {
+    let mut config = Config::default();
+    config.mode = Mode::CompileFail;
+    config.src_base = PathBuf::from("tests/compile_fail");
+    config.target_rustcflags = Some("-L ../target/debug -L ../target/debug/deps".into());
+    // config.link_deps();
```
jamiebrynes7: Commented out code

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,662 @@
...
+};
+use spatialos_sdk_sys::worker::*;
+use std::{collections::BTreeMap, marker::PhantomData};
+
+pub type FieldId = u32;
+
+#[derive(Debug)]
+pub struct SchemaComponentUpdate<'a> {
+    pub(crate) internal: *mut Schema_ComponentUpdate,
+    fields: SchemaObject<'a>,
```
jamiebrynes7: Is there a reason that this one is special and we inline the `SchemaObject`? Whereas the other `Schema*` types we don't do a similar thing.

Also this is missing events :wink:

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,662 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct SchemaObject<'a> {
+    internal: *mut Schema_Object,
+    _marker: PhantomData<&'a ()>,
+}
+
+impl<'a> SchemaObject<'a> {
```
jamiebrynes7: No `index` or `count` methods?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,662 @@
...
+
+    fn index_field(object: &SchemaObject, field: FieldId, index: u32) -> Self::RustType {
+        let raw = unsafe { Schema_IndexBool(object.internal, field, index) };
+        raw != 0
+    }
+}
+
+impl<T: IndexedField> Sealed for Option<T> {}
+
+impl<T: IndexedField> SchemaField for Option<T> {
```
jamiebrynes7: Might not be important here specifically, but for future reference.. 

Important to note `Option<T>` in Rust != `option<t>` in schemalang. In schemalang `option` has reference semantics, not value semantics.

This would translate strictly to `Option<Box<T>>` but we likely will want to optimize for this (as we can detect cycles in the schema structure that would require reference semantics). 

---

Example: 

This is perfectly valid schema. 
```
type Bar {
	Foo field =  1;
}

type Foo {
	option<Bar> recursive = 1;
}
```

Generating Rust code 1:1 would result in: 

```rust
pub struct Bar {
	pub field: Foo,
}

pub struct Foo {
	pub recursive: Option<Bar>,
}
```

which will fail to compile unless you `Box` or otherwise indirect the inner type of `recursive`.

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,662 @@
...
+where
+    K: IndexedField,
+    V: IndexedField,
+    K::RustType: Ord,
+{
+    type RustType = BTreeMap<K::RustType, V::RustType>;
+
+    fn add_field(object: &mut SchemaObject, field: FieldId, map: &Self::RustType) {
+        // Create a schema object at the specified field in `object`.
+        let object: SchemaObject = unsafe { Schema_AddObject(object.internal, field) }.into();
```
jamiebrynes7: This still isn't quite right, you don't need this line. You just add key-value pair objects at the field id in the parent map.

jamiebrynes7: Similarly for the `get_field` implementation

randomPoison: Oh, of course 🤦‍♀️ I'll get this right some day.

spatialos-sdk/src/worker/schema.rs
randomPoison: Ah, that's a good catch! I'll add a test case for this in the example schema file. I'll probably start off by having `option<T>` always generate `Option<Box<T>>`, and we can implement an optimized version at some point that skips boxing when not necessary.

spatialos-sdk/src/worker/schema.rs
randomPoison: Nah, they're not necessary here. The methods on `SchemaObject` are extra sugar to simplify the generated code for the `SchemaObjectType` implementations (they remove the need to bring the specific traits into scope). The process of indexing into fields is always handled by the `SchemaField` impl for the field's type, so it doesn't need to be exposed on `SchemaObject` directly. You can see an example of this in `manual_serialization.rs`.

spatialos-sdk/src/worker/schema.rs
randomPoison: The answer to both of those is "Work In Progress" 🤷‍♀️

randomPoison:
I'm doing another major pass in order to make our usage of the various schema data types completely safe and correct. I'm going to close this PR and open a new one once I've got a working version together so I can get more feedback.

PR #90, opened 2019-04-06
Add CLI and SDK downloading to cargo-spatial

### LINKED ISSUE #74: Download the SpatialOS SDK as part of cargo-spatial
Users need to have the SDK downloaded before they can compile the Rust SDK, and this is currently handled as a manual step before they can even build their project. We should handle this automatically as part of `cargo-spatial`. This should include specifying a library version in your `Spatial.toml` configuration file.

### LINKED ISSUE #75: Download the spatial CLI as part of cargo-spatial
We currently require that the user install the `spatial` CLI tool separately in order to be able to use `cargo-spatial`. We should reasonably be able to automate this in `cargo-spatial`, automatically downloading the CLI and installing it to a local directory if the user doesn't have it installed already.

jamiebrynes7:
Added the following commands to `cargo-spatial`: 

* `cargo spatial download cli` - this will download and run the installer. This is only supported on Windows/MacOS systems.
* `cargo spatial download sdk` - this downloads the required libraries as specified in the `Spatial.toml` file. 

If you provide a `--sdk-version` argument & have a `SPATIAL_LIB_DIR` env variable set, we don't error on not having a `Spatial.toml` config. This us to not break our own workflow in the repository.

Removed the `download_sdk` binary in `spatialos-sdk-tools` crate. That crate could _probably_ go away in the future, as the `generate_bindings` could be replaced with a shell script likely. 

**Unresolved Questions**
- [ ] We can implement a separate process for installing the CLI on Linux where we download the binary directly and `chmod +x` it. However, where do we put it? Ask users? Conditional arguments depending on OS? 
- [ ] Currently we force you to download the C API for Windows, Linux, and Mac as well as the tooling for your system. We could make this more granular in the future, through a configuration file. 

Fixes #74 and fixes #75


jamiebrynes7:
Also need to test ```cargo spatial download cli``` on a MacOS machine. Windows works provided you run it from a shell with the correct permissions (or have UAC disabled?)

cargo-spatial/src/download.rs
```diff
@@ -0,0 +1,256 @@
...
+            SpatialToolsPackage::SnapshotConverterLinux => "snapshot_converter-x86_64-linux",
+        }
+    }
+
+    fn get_relative_target_directory(&self) -> &str {
+        match self {
+            SpatialToolsPackage::SchemaCompilerWin
+            | SpatialToolsPackage::SchemaCompilerMac
+            | SpatialToolsPackage::SchemaCompilerLinux => "schema_compiler",
+            SpatialToolsPackage::SnapshotConverterWin
```
randomPoison: So I *think* Rust supports a leading `|` character in match branches for situations like this, so you should be able to do:

```
| SpatialToolsPackage::SchemaCompilerWin
| SpatialToolsPackage::SchemaCompilerMac
| SpatialToolsPackage::SchemaCompilerLinux => "schema_compiler",
```

That said, I don't know if it's actually considered better style to do it that way (I think that was added to make code generation easier), so defer to rustfmt if it tells you to not have the leading `|`.

cargo-spatial/src/download.rs
```diff
@@ -0,0 +1,256 @@
...
+    fn get_relative_target_directory(&self) -> &str {
+        match self {
+            SpatialWorkerSdkPackage::CApiWin => "win",
+            SpatialWorkerSdkPackage::CApiMac => "macos",
+            SpatialWorkerSdkPackage::CApiLinux => "linux",
+        }
+    }
+}
+
+enum SpatialToolsPackage {
```
randomPoison: Would be good to derive some common traits here, e.g. `derive(Debug, Clone, Copy, PartialEq, Eq, Hash)`. Same goes for all the other enums you define here.

cargo-spatial/src/download.rs
```diff
@@ -0,0 +1,256 @@
+// Conditionally re-export the correct version of download_cli dependent on platform.
+#[cfg(target_os = "linux")]
+pub use self::linux::*;
+#[cfg(any(target_os = "windows", target_os = "macos"))]
+pub use self::win_osx::*;
+
+use crate::config::Config;
+use crate::opt::DownloadSdk;
```
randomPoison: You could combine these two `use crate::` statements.

cargo-spatial/src/codegen.rs
```diff
@@ -12,7 +12,9 @@ use tap::*;
 ///
 /// Assumes that the current working directory is the root directory of the project,
 /// i.e. the directory that has the `Spatial.toml` file.
-pub fn run_codegen(config: &Config) -> Result<(), Box<dyn std::error::Error>> {
+pub fn run_codegen() -> Result<(), Box<dyn std::error::Error>> {
+    let config = Config::load()?;
```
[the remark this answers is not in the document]
jamiebrynes7: The previous setup would error out if no `Spatial.toml` was present. For something like:

`cargo spatial download cli` I wouldn't expect a `Spatial.toml` to be present 

However I can adjust this back to taking `&Config` as a parameter and just inline the loading when its called from `main.rs`

cargo-spatial/src/config.rs
```diff
@@ -80,12 +86,18 @@ impl Config {
...
-            .clone()
-            .or_else(|| std::env::var("SPATIAL_LIB_DIR").ok())
+        match self.spatial_lib_dir {
+            Some(ref path) => {
+                let mut cwd = ::std::env::current_dir().unwrap();
+                cwd.push(path);
+
+                cwd.to_str().map(|p| p.to_owned())
+            }
+            None => ::std::env::var("SPATIAL_LIB_DIR").ok(),
```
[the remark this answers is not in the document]
jamiebrynes7: Yes that seems like a good idea - I'd expect these sorts of env variables to be absolute.

cargo-spatial/src/config.rs
```diff
@@ -80,12 +86,18 @@ impl Config {
...
+    /// Returns the absolute path to the spatial SDK directory, or `None` if the path hasn't
     /// been configured.
     pub fn spatial_lib_dir(&self) -> Option<String> {
-        self.spatial_lib_dir
-            .clone()
-            .or_else(|| std::env::var("SPATIAL_LIB_DIR").ok())
+        match self.spatial_lib_dir {
+            Some(ref path) => {
+                let mut cwd = ::std::env::current_dir().unwrap();
+                cwd.push(path);
```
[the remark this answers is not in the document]
jamiebrynes7: Yeah that's a good shout

cargo-spatial/src/download.rs
jamiebrynes7: This was run through `rustfmt`, but I'm not sure it would change 

```x | y | z```

to 

```
| x
| y
| z
```

I can test for science :wink:

jamiebrynes7:
Looks like Mac doesn't work. The spatial installer is a `pkg` and needs to be invoked in a special way _sigh_ 

Also may require full `sudo` level permissions

PR #95, opened 2019-04-19
Implement cleared fields

### LINKED ISSUE #84: Implement cleared fields on updates
Noticed that there is a TODO about cleared fields https://github.com/jamiebrynes7/spatialos-sdk-rs/blob/master/spatialos-sdk/src/worker/internal/schema.rs#L73

We need to implement this to enable clearing of lists, maps, and options in component updates.

jamiebrynes7:
Implemented cleared fields for updates. 

Made sure to only call `SchemaComponentUpdate::cleared_fields()` if there are actually fields to check for clearedness(?).

Added some test schema for each of the cases (option, list, and map) 

Fixes #84

### DIFF
--- a/spatialos-sdk-code-generator/src/generator.rs
+++ b/spatialos-sdk-code-generator/src/generator.rs
@@ -443,6 +443,34 @@ impl Package {
             deserialize_expr
         }
     }
+
+    fn get_cleared_data(&self, field: &FieldDefinition) -> &str {
+        if field.option_type.is_some() {
+            return "None";
+        }
+
+        if field.list_type.is_some() {
+            return "Vec::new()";
+        }
+
+        if field.map_type.is_some() {
+            return "BTreeMap::new()";
+        }
+
+        panic!("Called get_cleared_data on a field which cannot be cleared!")
+    }
+
+    fn get_empty_field_expr(&self, field: &FieldDefinition, field_name: &str) -> String {
+        if field.option_type.is_some() {
+            return format!("{}.is_some()", field_name);
+        }
+
+        if field.list_type.is_some() || field.map_type.is_some() {
+            return format!("{}.is_empty()", field_name);
+        }
+
+        panic!("Called get_empty_field_expr on a field which cannot be empty!")
+    }
 }
 
 #[derive(Debug)]
--- a/spatialos-sdk/src/worker/internal/schema.rs
+++ b/spatialos-sdk/src/worker/internal/schema.rs
@@ -9,7 +9,7 @@ pub type FieldId = u32;
 pub struct SchemaComponentUpdate {
     pub component_id: ComponentId,
     pub internal: *mut Schema_ComponentUpdate,
-    pub cleared_fields: Option<Vec<FieldId>>
+    pub cleared_fields: Option<Vec<FieldId>>,
 }
 
 #[derive(Debug)]
@@ -40,7 +40,7 @@ impl SchemaComponentUpdate {
         SchemaComponentUpdate {
             component_id,
             internal: unsafe { Schema_CreateComponentUpdate(component_id) },
-            cleared_fields: None
+            cleared_fields: None,
         }
     }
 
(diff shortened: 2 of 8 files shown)


spatialos-sdk-code-generator/src/generator.rs
```diff
@@ -443,6 +443,34 @@ impl Package {
...
         }
     }
+
+    fn get_cleared_data(&self, field: &FieldDefinition) -> &str {
+        if field.option_type.is_some() {
+            return "None";
+        }
+
+        if field.list_type.is_some() {
+            return "Vec::new()";
```
randomPoison: I wonder if it would be a good idea to use the [`clear`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.clear) method to clear collections, instead of creating an entirely new one. Doing so would mean that the collection's capacity is retained, avoiding extra allocations if the user adds new elements to them. On the other hand, if the component in question is read-only in the worker (or simply doesn't get modified frequently), I could imagine a pathological case where you could end up with a bunch of collections in components that have extra capacity that's never used, resulting in excessive allocated memory 🤔 

How do you handle it in the C++ SDK? Do you completely clear the collection, including its allocated capacity?

jamiebrynes7: The collection in the update struct doesn't actually exist yet. For example, given the following schema:

```
component Foo {
	id = 1000;
	list<float> data = 1;
}
```

Where the generated Rust update struct is: 

```rust
pub struct FooUpdate {
	pub data: Option<Vec<f32>>
}
```

The cleared fields API allows you to determine whether the field 1 was cleared, or just wasn't updated. 

If you didn't have the cleared fields you cannot distinguish between an empty option of a list (not updated) or an option of an empty list (cleared) as they are both represented as 0 elements serialized at field id = 1. The cleared fields API allows you to distinguish between these two cases.

Taking this into Rust-land, after deserializing the update and _before_ processing the cleared fields your update struct will look like the following for *both* a list that wasn't updated, and a list that was cleared: 

```rust
FooUpdate {
	data: None
}
```

Then you process the cleared fields and set the field to the "empty field" equivalent if the field was cleared: 

```rust
FooUpdate { 
	data: Some(Vec::new())
}
```

---

You could allow the cleared fields API to leak into the Rust component update API, like: 

```rust
pub enum ClearableField<T> {
	Updated(T),
	Cleared,
	NotUpdated
} 
```

Such that the update struct looks like: 

```rust
pub struct FooUpdate {
	pub data: ClearableField<Vec<f32>>
}
```

But I'm not sure how I feel about that :confused:. It feels a bit like a leaky abstraction

spatialos-sdk/src/worker/internal/schema.rs
```diff
@@ -70,7 +70,20 @@ impl SchemaComponentUpdate {
         }
     }
 
-    // TODO: Cleared fields.
+    pub fn cleared_fields(&self) -> Vec<FieldId> {
```
davedissian: Would it be better to expose an API like `is_field_cleared`?

jamiebrynes7: Yes - you could store the cleared fields as a `Option<Vec<FieldId>>` when the object is created, then query it.

If this aligns with how the cleared fields API (I recall talking about the current implementation with you) - then this 100% makes sense

spatialos-sdk-code-generator/src/generator.rs
randomPoison: Ah, I completely misunderstood which piece of the code I was looking at! Representing a cleared list as `Some(Vec::new())` is totally fine by me.

I was originally referring to the logic around merging the update into an existing instance of the component. In the case where the field is cleared, I was wondering if it makes sense to clear the existing `Vec` in the component's field, rather than replacing it with the new, empty `Vec`. Replacing it completely means that the old `Vec` is dropped and its capacity is deallocated, which is inefficient if the worker code is going to push more elements into the list later.

jamiebrynes7:
Needs to be updated with the new API introduced in `14.0.0`

PR #96, opened 2019-04-22
Implement enums in code generation

jamiebrynes7:
Noticed that enums weren't fully implemented in the code generation side. We now generate the elements of the enum along with a `From<u32>` impl to help with the deserialization side.


davedissian:
Can we bring back the generated code example? Makes it more difficult to review otherwise.

davedissian:
(whoops).

jamiebrynes7:
Yeah I'm happy to pull the generated code back in

PR #99, opened 2019-04-23
Actually fix entity id this time

jamiebrynes7:
Added a test case for it. One day I'll pull in the exhaustive schema test package.

### DIFF
--- a/project-example/schema/example.schema
+++ b/project-example/schema/example.schema
@@ -25,3 +25,9 @@ component Rotate {
     /// The radius of the rotation.
     double radius = 3;
 }
+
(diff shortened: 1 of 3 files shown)


randomPoison:
Oooooooo, is there an exhaustive schema test package? I've been gradually building up test cases as part of the codegen rewrite, but it would be nice if there was a pre-existing test package I could lean on as well.

jamiebrynes7:
Yes - it was released in the `13.7` release, so once we are updated to that we can grab it (although we could probably explicitly pull that version for that package as a workaround if we really wanted to)

PR #101, opened 2019-04-26
Serialization rewrite pt. 2 - This time we're onto something

randomPoison:
After SEVERAL WEEKS OF TRIAL AND ERROR I think I'm finally on the right track in terms of figuring out a safe abstraction for the schema data types. Rather than wrapping up the entire pointer (e.g. `struct SchemaObject(*mut Schema_Object);`), we create a custom dummy type and use pointer casts to convert between the C types and the Rust types. Here's a tweaked version of getting a schema object from a `Schema_ComponentData` to demonstrate:

```rust
impl ComponentData {
    pub fn fields(&self) -> &Object {
        // `&self` is secretly a `*mut Schema_ComponentData`, so we
        // cast back to the original pointer in order to use it with
        // the C API.
        let component_data_ptr = self as *mut Schema_ComponentData;

        // Call the C API function with the `*mut Schema_ComponentData`
        // and get a `*mut Schema_Object` back.
        let schema_object = Schema_GetComponentDataFields(component_data_ptr);
       
        // Cast the raw `*mut Schema_Object` to a pointer to our
        // custom `Object` wrapper type, then dereference it to get
        // a `&Object`. Because Rust is ~magic~, the lifetimes will
        // correctly match up such that the reference to the object
        // can't outlive the `Schema_ComponentData` that owns it.
        &*(schema_object as *const Object)
    }
}
```

This means that we can expose an idiomatic API, using `&T` and `&mut T` to control mutability and ownership like we would with any other Rust types, while still being 100% safe.

This works because all of the schema types are opaque, zero-sized handle types, so we can define our own custom handle types (with custom methods and junk) and it's completely valid so long as we never try to pass any of these types by value. To enforce this, we construct the API such that:

* We only ever give the user a reference to the type, such that the user can't move out of the reference to take the type by value.
* The user cannot create their own instance of the type by value (because all fields are private), ensuring that users cannot use things like [`mem::replace`](https://doc.rust-lang.org/std/mem/fn.replace.html) to move out of a reference to the type.

The one case that's slightly unusual is when the user actually owns the object (and is therefore responsible for destroying the object when they're done), as is the case with `Schema_CreateComponentData()`. For these cases, I've added a custom type `Owned` that acts as a smart pointer to the data. This is functionally equivalent to [`Box`](https://doc.rust-lang.org/std/boxed/struct.Box.html), but works with the SpatialOS-specific functions for freeing the objects when dropped. As such, functions like `ComponentData::new` return an `Owned<ComponentData>`, rather than directly returning a `ComponentData`.

Please let me know what you think of this approach 😁 I'm especially curious if there are any soundness holes that you can see, or if this approach incorrectly models the C API's semantics in any way.

---

Also, I've temporarily removed everything related to component updates, commands, and events. This was so that I would only have to focus on `Schema_ComponentData` and `Schema_Object` when figuring out how to build a safe, idiomatic API around the C API. I'll start restoring the other data types shortly.

---

Update 2019-06-15: I've re-added support for component updates, see #103 for the original PR discussing the changes.


dgav:
@randomPoison I have some free time in the next few days - is this ready for review?

randomPoison:
Not ready for full review, no. I still need to re-add commands, then update the existing code generation for the new APIs. If you want to give it a look over and give me any thoughts on the approach I'm taking, though, I would greatly appreciate it!

dgav:
OK, np. Let me know if I can assist at all. Will give it a look over.

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,606 @@
...
+
+impl<K, V> SchemaFieldSealed for BTreeMap<K, V>
+where
+    K: IndexedField,
+    V: IndexedField,
+    K::RustType: Ord,
+{
+}
+
+impl<K, V> SchemaField for BTreeMap<K, V>
```
dgavedissian: Instead of implementing it on an instantiation of `BTreeMap` with the marker types as keys and values, can we introduce something like `pub struct SchemaMap<K, V>;` as a kind of "marker map"?

I guess this would change the usage to something like:
```
data.fields().field::<SchemaMap<String, i32>>(2);
```
instead of
```
data.fields().field::<BTreeMap<String, i32>>(2);
```

spatialos-sdk/src/worker/schema/component_update.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    pub fn add_cleared(&mut self, field: FieldId) {
+        unsafe {
+            Schema_AddComponentUpdateClearedField(self.as_ptr(), field);
+        }
+    }
+
+    pub fn field_cleared(&self, field: FieldId) -> bool {
+        // TODO: Implement this using `Schema_IsComponentUpdateFieldCleared` once we update
```
dgavedissian: If you rebase this branch once I merge the 14.0 PR, you can use the `Schema_IsComponentUpdateFieldCleared` API.

spatialos-sdk/src/worker/schema/component_update.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        result
+    }
+
+    pub(crate) unsafe fn from_raw<'a>(raw: *mut Schema_ComponentUpdate) -> &'a Self {
+        &*(raw as *mut _)
+    }
+
+    pub fn component_id(&self) -> ComponentId {
+        unsafe { Schema_GetComponentUpdateComponentId(self.as_ptr()) }
```
dgavedissian: Note that `Schema_GetComponentUpdateComponentId` (and friends) is gone in 14.0.0. You should instead get the component ID (and command index for commands) from the oplist when the `Worker_ComponentUpdate`/`Worker_CommandRequest` is given to you, and passed along to the required functions. Perhaps `from_raw` can take a component ID and command index (where applicable).

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,606 @@
...
+pub trait SchemaField: Sized + SchemaFieldSealed {
+    type RustType: Sized;
+
+    fn get_field(object: &Object, field: FieldId) -> Self::RustType;
+
+    fn add_field(object: &mut Object, field: FieldId, value: &Self::RustType);
+
+    fn has_update(update: &ComponentUpdate, field: FieldId) -> bool;
+
+    fn get_update(update: &ComponentUpdate, field: FieldId) -> Option<Self::RustType> {
```
dgavedissian: I find `has_update`, `get_update` and `add_update` to be quite confusing at first. Perhaps they should be called `update_has_field`, `update_get_field` and `update_add_field`?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -0,0 +1,606 @@
...
+
+pub use self::{component_data::*, component_update::*, object::*};
+
+pub type FieldId = u32;
+
+// =================================================================================================
+// Schema Conversion Traits
+// =================================================================================================
+
+pub trait SchemaField: Sized + SchemaFieldSealed {
```
dgavedissian: Nitpick: Can you remove the spacing between trait functions?

dgavedissian:
Overall, I love how this looks! Nice job.

randomPoison:
This refactoring has become a quagmire and I haven't had the time to work on it in a while. I'm going to close this ticket and try building out a plan to redo this rewrite piecemeal, such that it doesn't have to be  merged all at once. I'll leave the branch up for the time being, though, so that we can reference the first attempt as necessary.

PR #102, opened 2019-04-30
Add user handle abstraction

randomPoison:
I've added a new `UserHandle` type that handle the work of freeing the handle once we're done with it. It's meant to make it harder for use introduce memory leaks (such as the ones addressed in #93), while also making it easier to safely manage user handles (as `Entity` needs to do in order to support non-vtable serialization). I've still kept the existing free functions, since they're useful for the vtable functions (though I've renamed them to more accurately reflect their new role).

### DIFF
--- a/spatialos-sdk/src/lib.rs
+++ b/spatialos-sdk/src/lib.rs
@@ -1,5 +1,8 @@
 extern crate spatialos_sdk_sys;
 
+// TODO: Where should this live? We only need it for tests in order to reduce
+// boilerplate, but it needs to live in the crate because we use it for both
+// internal and external tests.
 #[macro_export]
 macro_rules! dummy_component {
     ($component:ident, $update:ident) => {
--- a/spatialos-sdk/src/lib.rs
+++ b/spatialos-sdk/src/lib.rs
@@ -1,4 +1,34 @@
 extern crate spatialos_sdk_sys;
 
+#[macro_export]
+macro_rules! dummy_component {
+    ($component:ident, $update:ident) => {
+        impl $crate::worker::schema::SchemaObjectType for $component {
+            fn from_object(_: &$crate::worker::schema::Object) -> Self {
+                unimplemented!()
+            }
+
+            fn into_object(&self, _: &mut $crate::worker::schema::Object) {
+                unimplemented!();
+            }
+        }
+
+        impl $crate::worker::component::Component for $component {
+            const ID: $crate::worker::component::ComponentId = 1234;
+            type Update = $update;
+        }
+
+        inventory::submit!($crate::worker::component::VTable::new::<
+            $component,
+        >());
+
+        pub struct $update;
+
+        impl $crate::worker::component::Update for $update {
+            type Component = $component;
+        }
+    };
+}
+
 pub(crate) mod ptr;
 pub mod worker;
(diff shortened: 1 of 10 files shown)


spatialos-sdk/src/worker/entity_builder.rs
```diff
@@ -57,7 +57,7 @@ impl EntityBuilder {
         Ok(())
     }
 
-    pub fn add_handle<C: Component, T: Into<String>>(
+    pub fn add_handle<C: Component + Send, T: Into<String>>(
```
jamiebrynes7: What's the use case for exposing this in addition to `add` on the builder?

spatialos-sdk/src/lib.rs
```diff
@@ -1,4 +1,37 @@
 extern crate spatialos_sdk_sys;
 
+// TODO: Where should this live? We only need it for tests in order to reduce
```
jamiebrynes7: Maybe in the internal module?

spatialos-sdk/src/worker/handle.rs
```diff
@@ -0,0 +1,132 @@
+//! A type-erased [`Arc`], used to pass user-defined data types to the C API.
```
jamiebrynes7: Should this class ever be exposed to user code? It feels like implementation detail and a leaky abstraction if it is.

Currently its a `pub mod` - https://github.com/jamiebrynes7/spatialos-sdk-rs/blob/user-handle-rewrite/spatialos-sdk/src/worker/mod.rs#L9

Perhaps it better belongs in the `internal` mod?

spatialos-sdk/src/lib.rs
jamiebrynes7: Actually scratch that, missed the part about external tests :confused:

spatialos-sdk/src/worker/entity_builder.rs
randomPoison: This is part of the larger story for allowing users more control over when serialization happens (i.e. whether to serialize immediately or to use a vtable). The idea is that the `add` function will "do the right thing" based on your global configuration (i.e. use vtables if vtables are enabled, serialize immediately if not), and then you have `add_handle` and `add_serialized` to manually override that global configuration. This doesn't fully address the needs of users who want to work without vtable serialization, but I think it's a necessary part of supporting those use cases.

spatialos-sdk/src/lib.rs
randomPoison: Yeah, needing it both internally and externally causes problems 😬 Since it's going to be a while before any of this gets merged into master anyway, I was planning on punting on this for now (unless you have a good suggestion on how to handle it).

spatialos-sdk/src/worker/handle.rs
randomPoison: Ah, you're right, `UserHandle` should be entirely private to the SDK. I don't think there's any use case where exposing it to the user makes sense. I was thinking of switching it a `pub(crate)` module. Is there a strong preference between that or moving it to `internal`? Personally I'm not sure if having a separate `internal` module is necessary, though I'm not opposed to moving `handle` into `internal` if you have strong preference that way 🙂

randomPoison:
Since it seems like this is generally agreed upon as being a useful abstraction, and all the outstanding issues are part of larger issues that will need to be addressed as part of the main refactoring PR, I'm going to close this PR and get it merged into the `serialization-refactor` branch. Thanks for the feedback!

PR #103, opened 2019-06-04
Component update rewrite

randomPoison:
I've finished implementing component updates for the serialization rewrite. This continues to follow the new patterns for handling schema data, and attempts to encode the semantics for component updates within Rust's type system.

* Added `schema::ComponentUpdate` type to represent `Schema_ComponentUpdate`.
* The `component::Update` trait defines the functions for serializing updates in order to reflect the fact that the top-level component update gets diffed, but any sub-objects do not.
* `get_update` and `add_update` have been added to the `SchemaField` trait (rather than creating a separate trait for updates) since all schema field types must be able to be added to a component update.
* Added support for sending/receiving component updates without using the vtable for that component type.
* Centralized most of the raw schema object access into methods on `Object`, so now there's almost no direct access to the C API in `schema` (just for implementing the primitive types via macro).


spatialos-sdk/src/worker/schema.rs
```diff
@@ -1,47 +1,59 @@
...
 }
 
+/// A type that can be efficiently added to a schema object as a list.
+///
+/// Since types that are serialized/deserialized as a list are copied directly into
+/// the schema object, any types that implement this trait must be ABI-compatible
+/// with the underlying C API. This means that only primitive types (including
+/// [`EntityId`]) implement `ArrayField`.
+///
+/// [`EntityId`]
```
jamiebrynes7: Is this line of the docs intended?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -414,29 +446,41 @@ where
...
+
+    fn has_update(update: &ComponentUpdate, field: FieldId) -> bool {
+        update.fields().object_field_count(field) > 0
+    }
+
+    fn add_update(update: &mut ComponentUpdate, field: FieldId, value: &Self::RustType) {
+        if !value.is_empty() {
+            Self::add_field(update.fields_mut(), field, value);
+        } else {
+            update.add_cleared(field);
```
jamiebrynes7: This is a nice way of handling the cleared field stuff 👍

spatialos-sdk/src/worker/schema/component_update.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub(crate) unsafe fn from_raw<'a>(raw: *mut Schema_ComponentUpdate) -> &'a Self {
+        &*(raw as *mut _)
+    }
+
+    pub fn component_id(&self) -> ComponentId {
+        unsafe { Schema_GetComponentUpdateComponentId(self.as_ptr()) }
+    }
+
+    pub fn deserialize<U: Update>(&self) -> U {
+        // TODO: Does it make sense to assert here? Or should we allow for runtime recovery
```
jamiebrynes7: Depends who calls this - if its library code exclusively, then I'd say assert as it indicates a critical flaw in the library.

If this is a user-facing API, then I'd say return a `Result<U, SomeError>`

spatialos-sdk/src/worker/schema/component_update.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    pub fn field_array<T: ArrayField>(&self, field: FieldId) -> Option<Vec<T::RustType>> {
+        if T::has_update(self, field) {
+            Some(self.fields().field_array::<T>(field))
+        } else {
+            None
+        }
+    }
+
+    pub fn add_field_array<T: ArrayField>(&mut self, field: FieldId, value: &[T::RustType]) {
```
jamiebrynes7: Should this signature be something like: 

```rust
pub fn add_field_array<T: ArrayField, U: AsRef<[T::RustType]>(&mut self, field: FieldId, value: &U)
```

so we can be a bit more generous with what we take in?

spatialos-sdk/src/worker/schema/component_update.rs
```diff
@@ -0,0 +1,159 @@
...
+    type Raw = Schema_ComponentUpdate;
+
+    unsafe fn destroy(inst: *mut Self::Raw) {
+        Schema_DestroyComponentUpdate(inst);
+    }
+}
+
+unsafe impl Send for ComponentUpdate {}
+
+struct ClearedFieldIter<'a> {
```
jamiebrynes7: When we upgrade to 13.8, there's a additional API  call for interacting with cleared fields which will make this a little easier (https://docs.improbable.io/reference/13.8/releases/release-notes#worker-sdk-in-c) 

For example - instead of calling `field` for each field on the object then iterating through cleared fields, the `field` could check itself for "cleared-ness" and return the appropriate object

spatialos-sdk/src/worker/connection.rs
```diff
@@ -487,6 +495,30 @@ impl Connection for WorkerConnection {
         }
     }
 
+    fn send_component_update_serialized(
```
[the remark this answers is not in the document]
davedissian: That's not correct. The C API will only call the vtable serialize method iff `schema_type` is `nullptr`. The function which calls the vtable serialize function (inside the C API) looks roughly like:

```
OutgoingData ManagedClientHandle::Serialize() && {
  auto* serialize = get_serialize_function(vtable)
  if (serialize && !schema_type) {
    SchemaDataType* serialized_schema_type = nullptr;
    serialize(component_id, vtable.user_data, user_handle.get(), &serialized_schema_type);
    ...
  }
  ...
}
```

so this shouldn't segfault in that case.

However, it _would_ segfault if the user provides null for both `schema_type` and `user_handle` (in the `Worker_ComponentUpdate`).

davedissian:
Will review this properly sometime this weekend :)

spatialos-sdk/src/worker/connection.rs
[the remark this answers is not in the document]
jamiebrynes7: Awesome :smile:

spatialos-sdk/src/worker/component.rs
```diff
@@ -115,28 +118,35 @@ impl<'a> ComponentDataRef<'a> {
...
+    pub user_handle: *mut Worker_ComponentUpdateHandle,
+}
 
-//     // NOTE: `user_handle` is borrowing data owned by the parent object, but it's a
-//     // type-erased pointer that may be null, so we just mark that we're borrowing
-//     // *something*.
-//     pub _marker: PhantomData<&'a ()>,
-// }
+impl<'a> ComponentUpdateRef<'a> {
+    pub unsafe fn from_raw(data: &'a Worker_ComponentUpdate) -> Self {
```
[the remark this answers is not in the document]
randomPoison: I had a couple of reasons for moving away from the `From` impl:

* The main reason is that the conversion function needs to be marked as unsafe. Since the wrapper type is meant to be a safe abstraction over the C API, its safety depends on it being correctly constructed from valid data.
* Trait impls are public, whereas it's important that the conversions be private to the crate in order to enforce safety (by ensuring that users can never construct an invalid object).
* `from_raw` more accurately reflects the nature of the function, that it's a conversion from the raw C representation into a safe Rust representation. This is a more minor point, but it feels like `From` isn't the correct idiom in this case, so an explicit trait method seems to me like the better approach.

spatialos-sdk/src/worker/component.rs
```diff
@@ -295,6 +305,8 @@ unsafe extern "C" fn vtable_component_data_deserialize<C: Component>(
     let schema_data = schema::ComponentData::from_raw(schema_data);
     let component = schema_data.deserialize::<C>();
     *handle_out = handle::allocate_raw(component);
+
+    // TODO: Do we need to do proper error handling support here?
```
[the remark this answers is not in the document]
randomPoison: I'm not entirely sure what sort of errors it makes sense to handle. The deserialization logic previously returned a `Result`, so it used to be that we would indicate an error if it returned `Err`. However, the underlying C API seems to (somewhat) hide the potential error cases (like missing fields), so it's not clear to me what errors make sense to handle here. My two questions would be:

* How is this handled in other language SDKs?
* What is the purpose for allowing error handling in the deserialization function (@davedissian might be able to give examples here).

spatialos-sdk/src/worker/connection.rs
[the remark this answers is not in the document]
randomPoison: Yup yup! I've started to add methods like this, where the user can specifically choose to opt out of vtable serialization. This doesn't fully address the needs of users who want to avoid vtable serialization, but it's a first step in that direction.

spatialos-sdk/src/worker/schema.rs
randomPoison: Ah, my mistake! I meant to make that a link to the relevant docs page, but must have gotten distracted 😅

spatialos-sdk/src/worker/schema/component_update.rs
randomPoison: Ah, that's a good point! I'll have a look at how this function is used and whether or not it needs to be user-facing. I think right now it's only used internally within `ComponentUpdateOp`, but if there's a reasonable use case for users to want to call it, then I'll have it return a `Result` 😁

spatialos-sdk/src/worker/schema/component_update.rs
randomPoison: That's terrific! That's what I had originally envisioned doing when I started writing the API, and had to switch to the iterator-based approach you see in `manual_serialiation.rs` to avoid O(n^2) performance when deserializing component updates. Do you know if `Schema_IsComponentUpdateFieldCleared` is guaranteed to be a constant-time lookup? If it's just walking through the list under the hood, then it's not much of an improvement, but if it's actually doing an array lookup (or something similarly efficient) then we should absolutely use it!

davedissian: Yep guaranteed to be constant time.

spatialos-sdk/src/worker/schema/component_update.rs
randomPoison: I'm not sure that's necessary in this case. It's like taking `&str` vs `AsRef<str>`. In most cases I lean towards taking `&str` or `&[T]` for simplicity, and only using generics when there's a specific, known use case that would benefit from doing so. That said, it doesn't seem like there's any technical reason why we couldn't use the generic version here, so I'm okay with switching to that if you think it's better :)

spatialos-sdk/src/worker/schema/component_update.rs
randomPoison: I opted to return an `Option<U>` since that's what I'm doing in similar scenarios 👍

spatialos-sdk/src/worker/schema/component_update.rs
randomPoison: I've added a `field_cleared` method to `ComponentUpdate`. Right now it's doing a linear search in the cleared fields list, but we can easily switch to using `Schema_IsComponentUpdateFieldCleared` once we update the SDK. That nicely simplified the code we need to generate for field updates 🎉

spatialos-sdk/src/worker/schema/component_update.rs
jamiebrynes7: No strong use case but gives a bit more flexibility to the caller. That being said - if it implements `AsRef` then the caller can just do `foo.as_ref()` at the call site. 🤷‍♂

spatialos-sdk/src/worker/schema.rs
```diff
@@ -413,7 +413,18 @@ impl<T: IndexedField> SchemaField for Option<T> {
...
     fn has_update(update: &ComponentUpdate, field: FieldId) -> bool {
-        T::field_count(update.fields(), field) > 0
+        T::field_count(update.fields(), field) > 0 || update.field_cleared(field)
+    }
+
+    fn get_update(update: &ComponentUpdate, field: FieldId) -> Option<Self::RustType> {
+        if update.field_cleared(field) {
+            Some(None)
+        } else {
+            match Self::get_field(update.fields(), field) {
```
jamiebrynes7: Little bit of sugar...

```rust
Self::get_field(update.fields(), field)
	.map(|opt| Some(opt))
```

randomPoison: Ah, good catch! I felt like I was missing a better way of writing that 🤦‍♀

randomPoison: Also, even better, you can do `.map(Some)`, because `Some` (other tuple-structs) can be used as functions ✨

spatialos-sdk/src/worker/component.rs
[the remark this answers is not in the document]
randomPoison: I'm going to merge the PR into the main rewrite branch as-is, but I think this point is worth looking at more, either as part of the main PR or continuing the discussion here.

PR #104, opened 2019-06-16
Add events to component updates (serialization rewrite)

randomPoison:
Since component events and fields can't have the same name, we can add events directly to the generated component update type, which makes implementing events pretty trivial 😁

### DIFF
--- a/spatialos-sdk/tests/manual_serialization.rs
+++ b/spatialos-sdk/tests/manual_serialization.rs
@@ -12,7 +12,7 @@
 //!     NestedType nested = 6;
 //!     list<bool> events = 7;
 //!
-//!     event NestedType nested;
+//!     event NestedType nested_event;
 //!     event CoolEvent cool_event;
 //! }
 //!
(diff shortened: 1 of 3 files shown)


spatialos-sdk/tests/manual_serialization.rs
```diff
@@ -1,10 +1,38 @@
+//! Example code for the following schema defintions:
+//!
+//! ```schema
+//! component CustomComponent {
```
davedissian: I think we should aim to generate the test code (rather than hand write it) in general.

randomPoison: Agreed! I think it's worth having one hand-written test-case, at least for now, for a couple of reasons:

* For now, it's helping me verify the serialization API that I'm creating, to make sure that it's actually *possible* for us to generate the necessary code using the public API.
* Once I'm updating the code-generation logic, it'll be good to have a hand-written example of what the generated code should look like in order to understand what the code generation needs to produce.

But I don't think we should bother with more than the one hand-written test case, and maybe we can drop it once I get around to updating the code generation for the new API 🙂

PR #105, opened 2019-07-18
Upgrade to SpatialOS 14.0.0

dgavedissian:


dgav:
Still need to expose the modular UDP connection parameters properly.

spatialos-sdk/src/worker/locator.rs
```diff
@@ -43,51 +75,69 @@ impl Drop for Locator {
 }
 
 pub struct LocatorParameters {
-    pub project_name: CString,
+    pub project_name: Option<CString>,
```
jamiebrynes7: What does a `project_name` of `None` mean?

spatialos-sdk/src/worker/locator.rs
```diff
@@ -264,3 +370,347 @@ pub(crate) extern "C" fn queue_status_callback_handler(
...
+        self.use_insecure_connection = true;
+        self
+    }
+
+    fn to_worker_sdk(&self) -> Worker_Alpha_PlayerIdentityTokenRequest {
+        Worker_Alpha_PlayerIdentityTokenRequest {
+            development_authentication_token: self.dev_auth_token.as_ptr(),
+            player_id: self.player_id.as_ptr(),
+            duration_seconds: match self.duration_seconds {
+                Some(ref value) => value as *const _ as *mut u32, // TODO: Remove double cast when C SDK is fixed.
```
jamiebrynes7: This TODO might be able to be resolved now 😄

spatialos-sdk/src/worker/snapshot.rs
```diff
@@ -18,38 +18,36 @@ impl SnapshotOutputStream {
...
-
-        let stream = SnapshotOutputStream { ptr };
-
-        let err_ptr = unsafe { Worker_SnapshotOutputStream_GetError(ptr) };
-        if !err_ptr.is_null() {
-            unsafe { Worker_SnapshotOutputStream_Destroy(ptr) };
-            return Err(cstr_to_string(err_ptr));
+        let state = unsafe { Worker_SnapshotOutputStream_GetState(stream_ptr) };
+        match Worker_StreamState::from(state.stream_state) {
+            Worker_StreamState_WORKER_STREAM_STATE_GOOD => {
```
jamiebrynes7: Are the error codes that this can return meaningful?

jamiebrynes7: If so - should we create an actual error enum to return here (and elsewhere)?

jamiebrynes7:
> Still need to expose the modular UDP connection parameters properly. Perhaps it's worth refactoring the whole thing to not expose RakNet or non-modular KCP as these are replaced by the much superior modular UDP network stack.

I'd be fine with this :smile: I feel like that's a reasonable opinion to implement. We probably will end up doing the same in the GDK in the near future.

spatialos-sdk/src/worker/connection.rs
```diff
@@ -218,7 +217,7 @@ impl WorkerConnection {
         WorkerConnectionFuture::new(future_ptr)
     }
 
-    pub fn connect_locator_async(
+    pub fn connect_locator_and_queue_async(
```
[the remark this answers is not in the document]
randomPoison: I'd say remove it, no point in deprecating when we don't have a stable API for the Rust SDK.

spatialos-sdk/src/worker/component.rs
```diff
@@ -425,6 +423,7 @@ unsafe extern "C" fn vtable_component_update_serialize<C: Component>(
 }
 
 unsafe extern "C" fn vtable_command_request_free<C: Component>(
+    _: u32,
```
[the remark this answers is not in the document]
dgavedissian: command request and response vtable functions now also pass along the `command_index`.

spatialos-sdk/src/worker/locator.rs
dgavedissian: It's unnecessary if you use the player authentication flow.

spatialos-sdk-code-generator/src/generator.rs
```diff
@@ -100,58 +85,74 @@ impl Package {
...
+            let data_type = self.resolve_type_reference(&data_definition);
+            data_type.fields.clone()
         } else {
-            component.field_definitions.clone()
+            component.fields.clone()
         }
     }
 
-    fn generate_value_type_reference(&self, value_type: &ValueTypeReference) -> String {
+    fn generate_rust_type_name(&self, value_type: &TypeReference) -> String {
```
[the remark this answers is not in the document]
dgavedissian: Might be a bit tricky to get serde to play ball here, as the whole point is for `TypeReference` to match the proto file.

spatialos-sdk-code-generator/Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
 version = "0.1.0"
-authors = ["David Avedissian <git@dga.me.uk>"]
+authors = ["David Avedissian <git@dga.dev>"]
 
 [dependencies]
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
-t4rust-derive = { git = "https://github.com/ReSpeak/t4rust" }
+t4rust-derive = "0.1.3"
```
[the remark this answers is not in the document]
dgavedissian: Yeah, seems the author published it recently. Lucky :P

spatialos-sdk-code-generator/src/generator.rs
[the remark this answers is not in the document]
jamiebrynes7: I guess to map to the proto perfectly you'd have: 

```rust
pub enum TypeReference_ValueType {
	Primitive(PrimitiveType),
	Enum(String),
	Type(String)
}

pub struct TypeReference {
	pub value_type: TypeReference_ValueType,
}
```

Definitely not necessary, but will make the code generator easier to read imo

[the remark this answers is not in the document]
jamiebrynes7: Serde looks like it _may_ support this: https://serde.rs/enum-representations.html

spatialos-sdk-code-generator/src/generator.rs
```diff
@@ -553,74 +537,43 @@ pub fn generate_code(bundle: SchemaBundle) -> String {
...
-            .borrow_mut()
-            .enums
-            .insert(qualified_name.clone(), enum_def);
-        package.enums.insert(qualified_name.clone());
-    }
-    /*
-    for pkg in &generated_code.borrow().packages {
-        println!("Found package: {}", pkg.qualified_name);
+        for type_def in file.types {
+            // TODO: handle outer type.
```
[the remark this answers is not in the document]
dgavedissian: Unsure, need to test it.

spatialos-sdk/src/worker/connection.rs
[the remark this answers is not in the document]
dgavedissian: OK. I'll delete the flow.

spatialos-sdk/src/worker/locator.rs
dgavedissian: I was able to remove the 2nd half of it (I guess the C API forgot to make it const)

spatialos-sdk-code-generator/src/generator.rs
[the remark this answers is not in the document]
dgavedissian: Done.

dgavedissian:
> I'd be fine with this smile I feel like that's a reasonable opinion to implement. We probably will end up doing the same in the GDK in the near future.

I ended up doing it. :P

spatialos-sdk/src/worker/snapshot.rs
dgavedissian: Done.

spatialos-sdk/src/worker/connection.rs
```diff
@@ -217,34 +217,6 @@ impl WorkerConnection {
...
-        let callback = Box::new(callback);
-        let callback_ptr = Box::into_raw(callback) as *mut ::std::os::raw::c_void;
-
-        unsafe {
-            let ptr = Worker_Locator_ConnectAndQueueAsync(
-                locator.locator,
-                deployment_name_cstr.as_ptr(),
-                &connection_params,
-                callback_ptr,
-                Some(queue_status_callback_handler),
```
jamiebrynes7: This `queue_status_callback_handler` can be removed as well now I think

spatialos-sdk-tools/Cargo.toml
```diff
@@ -9,6 +9,6 @@ name = "generate_bindings"
 path = "src/generate_bindings/main.rs"
 
 [dependencies]
-clap = "2.32.0"
-glob = "0.2"
-bindgen = "0.42"
+clap = "^2"
+glob = "~0.2"
```
jamiebrynes7: Any reason for relaxing the constraints on the versions here?

spatialos-sdk/Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2018"
 [dependencies]
 derivative = "1.0.2"
 spatialos-sdk-sys = { path = "../spatialos-sdk-sys"}
-futures = "0.1.25"
+futures = "0.1"
```
jamiebrynes7: Similarly here?

spatialos-sdk/src/worker/connection.rs
```diff
@@ -167,6 +168,55 @@ pub struct WorkerConnection {
...
+{
+    let mut conn_params = params.to_worker_sdk();
+
+    // conn_params.network will not have the protocol parameters setup (due to requiring pointers to things on the stack). Therefore, we do that here.
+    let mut udp_kcp_params = Worker_Alpha_KcpParameters::default();
+    let mut udp_erasure_codec_params = Worker_ErasureCodecParameters::default();
+    let mut udp_heartbeat_params = Worker_HeartbeatParameters::default();
+    let mut udp_flow_control_params = Worker_Alpha_FlowControlParameters::default();
+    match &params.network.protocol {
+        ProtocolType::Tcp(params) => {
```
jamiebrynes7: Small nit - can we not alias `params` from the function signature here? I was really confused when reading the match block

spatialos-sdk/src/worker/parameters.rs
```diff
@@ -218,45 +129,27 @@ impl NetworkParameters {
...
-        let (protocol_type, raknet_params, tcp_params, kcp_params, modular_udp_params) =
-            self.protocol_params.to_worker_sdk();
         Worker_NetworkParameters {
             use_external_ip: self.use_external_ip as u8,
-            connection_type: protocol_type,
-            raknet: raknet_params,
-            tcp: tcp_params,
-            kcp: kcp_params,
-            modular_udp: modular_udp_params,
+            connection_type: 0,
```
jamiebrynes7: Can we add an comment here saying that we will fill in these parameters after this call? To match up with the comments in `do_connect`

spatialos-sdk/src/worker/snapshot.rs
```diff
@@ -5,12 +5,39 @@ use crate::{
 use spatialos_sdk_sys::worker::*;
 use std::{ffi::CString, path::Path};
 
+#[derive(Debug)]
+pub enum SnapshotError {
+    Bad,
```
jamiebrynes7: Bad what? 😂

spatialos-sdk/src/worker/snapshot.rs
```diff
@@ -5,12 +5,39 @@ use crate::{
...
+    InvalidData,
+    EOF,
+}
+
+#[derive(Debug)]
+pub struct SnapshotErrorState {
+    pub error: SnapshotError,
+    pub message: String,
```
jamiebrynes7: Unsure if its more idiomatic to wrap this string into each of the error codes 🤔

spatialos-sdk-sys/build.rs
```diff
@@ -23,10 +23,13 @@ static PACKAGE_DIR: &str = "macos";
 static PACKAGE_DIR: &str = "win";
 
 fn main() {
+    /*
     let lib_dir = match env::var("SPATIAL_LIB_DIR") {
         Ok(s) => s,
         Err(_) => panic!("SPATIAL_LIB_DIR environment variable not set."),
     };
+    */
+    let lib_dir = "/home/dga/Projects/spatialos-sdk-rs/dependencies";
```
jamiebrynes7: 😛

spatialos-sdk-tools/Cargo.toml
dgavedissian: Nah not particularly. I'll revert it (as it can be done in a separate PR).

spatialos-sdk-sys/build.rs
dgavedissian: RIP.

spatialos-sdk/src/worker/snapshot.rs
dgavedissian: Yeah, it probably is... I'll change it.

spatialos-sdk/src/worker/snapshot.rs
dgavedissian: Well, this was supposed to be a "bad state". I'll change it to `BadState`

spatialos-sdk/src/worker/connection.rs
dgavedissian: Sure

spatialos-sdk/Cargo.toml
dgavedissian: Might as well follow the latest patch.

spatialos-sdk/src/worker/connection.rs
dgavedissian: I think I did delete it?

spatialos-sdk-code-generator/src/generator.rs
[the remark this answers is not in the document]
dgavedissian: Turns out it just works (tm).

dgavedissian:
Should be ready for review now.

randomPoison:
Overall I like the approach you ended up taking with regards to converting `ConnectionParameters` to `Worker_ConnectionParameters`. What you have is a lot better than the initial version using `Vec<Box<dyn Any>>` :+1: I'd be cool with this getting merged as-is.

That said, I'm not sure using a closure to get the various protocol parameters on the stack is the cleanest or safest approach. Instead, I'd suggest having an intermediate struct that extracts all of the C types out of `ConnectionParameters`, and then have a method on _that_ type that spits out a `Worker_ConnectionParameters` with pointers back into its data. It'd be a bit of boilerplate to setup, but I think it's a bit more of a direct solution that's ultimately easier to follow.

Would you be open to me making a PR against your branch to implement this? It might be easier to discuss the relative advantages/drawbacks of the two approaches if we could look at them both.

dgavedissian:
Sure, no problem. However, I'm unsure how this approach is less safe? The closure is there because the `Worker_ConnectionParameters` has pointers back to the stack, and the only safe way I can think of keeping those pointers in place (without code duplication) is to avoid clearing the stack frame by calling another function (specified by the caller of `do_connection`) to consume the data structure and return a connection future. However, I'd be curious to see your take on it.

randomPoison:
Ah yeah, "safe" isn't really the right word. There's not really a strong safety difference between the two as far as I can tell.

PR #106, opened 2019-07-23
WIP: Alternate ConnectionParameters to Worker_ConnectionParameters conversion

randomPoison:
Sketch out an alternate approach to converting a `ConnectionParameters` to a `Worker_ConnectionParameters`, specifically using a different strategy for populating the various pointers hiding in `Worker_Alpha_ModularUdpNetworkParameters `.

This approach uses a new enum `IntermediateProtocolType` to "flatten" the contents of `ProtocolType::Udp` into the corresponding C structs on the stack. From there, we can populate the final `Worker_ConnectionParameters` from the values on the stack.

From a safety perspective, this is functionally equivalent to the closure-based approach implemented in #105, which is to say that it's *mostly* safe but can't completely statically prevent misuse due to the nature of FFI and using C structs. I'd personally argue that the intermediate struct approach is a bit cleaner and easier to understand because it doesn't take over control flow the way that a closure does, however that's ultimately a minor point in this particular situation.

---

This is currently not entirely done, but there's enough in place to make the approach clear. I'll try to finish it up over the next couple of days, but let me know what you think in the meantime. I think its worth discussing both approaches, since neither is obviously better than the other.

### DIFF
--- a/spatialos-sdk/src/worker/parameters.rs
+++ b/spatialos-sdk/src/worker/parameters.rs
@@ -82,34 +82,8 @@ impl ConnectionParameters {
         }
     }
 
-    pub(crate) fn to_worker_sdk(&self) -> Worker_ConnectionParameters {
-        Worker_ConnectionParameters {
-            worker_type: self.worker_type.as_ptr(),
-            network: self.network.to_worker_sdk(),
-            send_queue_capacity: self.send_queue_capacity,
-            receive_queue_capacity: self.receive_queue_capacity,
-            log_message_queue_capacity: self.log_message_queue_capacity,
-            built_in_metrics_report_period_millis: self.built_in_metrics_report_period_millis,
-            protocol_logging: self.protocol_logging.to_worker_sdk(),
-            enable_protocol_logging_at_startup: self.enable_protocol_logging_at_startup as u8,
-            enable_dynamic_components: 0,
-            thread_affinity: self.thread_affinity.to_worker_sdk(),
-            component_vtable_count: if self.use_internal_serialization {
-                DATABASE.len() as u32
-            } else {
-                0
-            },
-            component_vtables: if self.use_internal_serialization {
-                DATABASE.to_worker_sdk()
-            } else {
-                ptr::null()
-            },
-            default_component_vtable: if self.use_internal_serialization {
-                ptr::null()
-            } else {
-                &vtable::PASSTHROUGH_VTABLE
-            },
-        }
+    pub(crate) fn flatten(&self) -> IntermediateConnectionParameters<'_> {
+        unimplemented!()
     }
 }
 
@@ -132,16 +106,6 @@ impl NetworkParameters {
             protocol: ProtocolType::Tcp(TcpNetworkParameters::default()),
             connection_timeout_millis: u64::from(WORKER_DEFAULTS_CONNECTION_TIMEOUT_MILLIS),
             default_command_timeout_millis: WORKER_DEFAULTS_DEFAULT_COMMAND_TIMEOUT_MILLIS,
-        }
-    }
-
-    pub(crate) fn to_worker_sdk(&self) -> Worker_NetworkParameters {
-        Worker_NetworkParameters {
-            use_external_ip: self.use_external_ip as u8,
-            connection_timeout_millis: self.connection_timeout_millis,
-            default_command_timeout_millis: self.default_command_timeout_millis,
-            // Other parameters will be set when establishing a connection.
-            ..Worker_NetworkParameters::default()
         }
     }
 }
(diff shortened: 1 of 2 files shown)


spatialos-sdk/src/worker/parameters.rs
```diff
@@ -430,3 +408,112 @@ impl ThreadAffinityParameters {
...
 }
+
+/// Helper struct for converting `ConnectionParameters` into `Worker_ConnectionParameters`.
+pub(crate) struct IntermediateConnectionParameters<'a> {
+    params: &'a ConnectionParameters,
+    protocol: IntermediateProtocolType,
+}
+
+impl<'a> IntermediateConnectionParameters<'a> {
+    pub(crate) fn as_raw(&self) -> Worker_ConnectionParameters {
```
jamiebrynes7: There's no guarantee that the `Worker_ConnectionParameters` lives as long as `&self`, so you can end up with dangling references here, right? 

Wouldn't it be more correct to construct the `Worker_ConnectionParameters` when you instantiate `IntermediateConnectionParameters` and only allow it to be borrowed? Thus tying the lifetimes of `ConnectionParameters` and the C representation? 

I.e. - 

```rust
pub(crate) struct IntermediateConnectionParameters<'a> {
    params: &'a ConnectionParameters,
    protocol: IntermediateProtocolType,
	raw_params: Worker_ConnectionParameters
}

impl<'a> IntermediateConnectionParameters<'a> {
	pub fn new(params: &'a ConnectionParameters, protocol: IntermediateProtocolType) -> Self{
		IntermediateConnectionParameters {
			params,
			protocol,
			Self::to_raw(params)
		}
	}

	pub fn raw(&self) -> &Worker_ConnectionParameters { &self.raw_params }

	fn to_raw(params: &ConnectionParameters) -> Worker_ConnectionParameters { ... }
}

Comment (randomPoison) [reply spatialos-sdk/src/worker/parameters.rs]: Problem is that `Worker_ConnectionParameters` is `Clone`/`Copy`, so it would still be trivial to create a copy of the `Worker_ConnectionParameters` that ends up outliving the `IntermediateConnectionParameters` that owns the data it points to.

As far as I can tell, there's no way to statically prevent misuse of `Worker_ConnectionParameters` (or any other C type with pointers in it). Since it uses raw pointers instead of references, it doesn't adhere to Rust's ownership systems and so the lifetimes are effectively invisible to the compiler. The best we can do is make something that is:

1. Easier to use correctly than not.
2. Easy to visually verify is correct when reviewing changes.

The one nice thing about this approach (as opposed to the closure-based approach) is that you never have to actually bind the `Worker_ConnectionParameters` to a variable. You can always pass `&intermediate.as_raw()` to the C API, which helps reduce the likelihood that you try to keep the `Worker_ConnecitonParameters` around longer than is valid. However, this is still only a convention and can't be enforced by the compiler 😞

Comment (jamiebrynes7) [reply spatialos-sdk/src/worker/parameters.rs]: I wonder if we can special case `Worker_ConnectionParameters` to not derive `Clone`/`Copy` on it. If so, we should be able to eliminate the risk (however low it may be)

Comment (jamiebrynes7) [reply spatialos-sdk/src/worker/parameters.rs]: There's many occurrences of this sort of thing throughout the codebase, it would be nice to be able to find a pattern that we can use to get as close to guaranteed safety as possible (up to stopping auto-derived traits for types, or even a blanket ban).

`Copy` in particular on a structure with pointers feels dangerous 😬

Comment (jamiebrynes7) [reply spatialos-sdk/src/worker/parameters.rs]: YOLO'ed together a trait-struct combo based on the above that we could try to use as a pattern if we had no `Copy/Clone` on the generated bindings. (No guarantee that it compiles 😉)

```rust
// Assuming we have no Copy/Clone on "U" the following struct and trait should provide a safe binding 
// between the Rust (safe) representation and the C (unsafe) representation.

pub(crate) trait WorkerSdkConversion<U> {
    fn raw(&Self) -> U;
}

pub(crate) struct RawBinding<'a, T, U> {
    safe: &'a T,
    raw: U
}

impl<'a, T, U> RawBinding<'a, T, U> 
    where T : WorkerSdkConversion<U>,
    /* U: !Copy + !Clone                if we had negative type constraints */
    {
        pub fn new(safe: &'a T) -> Self {
            Self {
                safe,
                raw: safe.raw()
            }
        }

        pub fn as_raw(&self) -> &U {
            &self.raw
        }
}

impl WorkerSdkConversion<Worker_ConnectionParameters> for ConnectionParameters {
    pub fn raw(&self) -> Worker_ConnectionParameters {
        ...
    }
}

impl WorkerConnection {
    pub fn make_connection(params: &ConnectionParameters) -> Future<WorkerConnection> {
        let binding = RawBinding::new(&params);
        let raw = binding.as_raw();

        drop(binding);
        connect(..., raw); // Invalid!
    }
}
```

randomPoison: > Of course, nothing prevents someone simply calling:
> 
> ```rust
> let unsafe_params = params.raw();
> ```

Yup 🙃 The problem is that FFI is inherently unsafe, and ultimately relies on some degree of convention and manual checking to enforce safety. We could setup `WorkerSdkConversion`/`RawBinding`, but even that's still just a convention, and there's nothing stopping us from just writing a regular `to_worker_sdk` method without going through the traits.

randomPoison:
@dgavedissian Thoughts? Feedback? Opinions?

dgavedissian:
Apologies, haven't got round to it, but I haven't forgotten :)

randomPoison:
No worries! I just don't want to keep your PR blocked if you're ready to get it moving along :+1:

spatialos-sdk/src/worker/parameters.rs
dgavedissian: Yeah, I agree with @randomPoison that it's probably not worth trying to guard against more specific scenarios, as this is really just a small bit of internal interop code.

spatialos-sdk/src/worker/parameters.rs
```diff
@@ -430,3 +408,112 @@ impl ThreadAffinityParameters {
...
+                Worker_NetworkParameters {
+                    connection_type:
+                        Worker_NetworkConnectionType_WORKER_NETWORK_CONNECTION_TYPE_TCP as u8,
+                    tcp,
+
+                    ..partial_network_params
+                }
+            }
+
+            IntermediateProtocolType::Udp {
```
dgavedissian: Why is there no & before `IntermediateProtocolType`?

randomPoison: In the above match arm, having the explicit `&` causes the `tcp` variable to be a copy of the value in `self.protocol`, wheras not having the `&` causes it to have the default behavior which is to borrow the data (which we want for the UDP case since we want pointers back into `self.protocol`.

I realize that it's pretty confusing in this case. Instead I can remove the `&` above and change it do `tcp: *tcp` to make the copy more explicit.

randomPoison:
Oh cool, glad you like it :grin:

spatialos-sdk/src/worker/parameters.rs
randomPoison: Made the change, hopefully that makes things a bit less confusing 🙂

spatialos-sdk/src/worker/parameters.rs
randomPoison: @jamiebrynes7 Were you satisfied with that explanation? Anything else you wanted to go over? If you're good with it, I can get this merged.

jamiebrynes7: Yeah all good with me! :smile:

PR #108, opened 2019-08-12
Checking in latest codegen

johnpmayer:
Also, how do you feel about checking in the files, generally?

Against checking-in, on principle version control systems really only ought to track human-generated content.

However, the idiosyncrasies of a build system sometimes lend to more convenience. For example, my workflow (VScode + RLS) fails on a clean checkout if I haven't run the cargo spatial codegen command. I'm not sure if this would work differently if the generated code were managed by a build.rs script. What good looks like, in my opinion, is [Pants Java Protobuf](https://www.pantsbuild.org/build_dictionary.html#bdict_java_protobuf_library) which seamlessly links generated code to your classpath (and IDE) without requiring the generated code to be checked-in.

### DIFF
--- a/project-example/src/generated.rs
+++ b/project-example/src/generated.rs
@@ -71,6 +71,38 @@ impl TypeConversion for CommandData {
 }
 
 #[derive(Debug, Clone)]
+pub struct TestType {
+    pub value: i32,
+}
+impl TypeConversion for TestType {
+    fn from_type(input: &SchemaObject) -> Result<Self, String> {
+        Ok(Self {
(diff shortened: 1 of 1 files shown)


randomPoison:
I actually have a (very) long-running refactor going on to make it so that code is generated in the `build.rs` for a crate in order to address this issue! Unfortunately, my work on that has slowed down due to a lack of personal time and my company moving away from looking at SpatialOS :cry: In the meantime, keeping the generated code in the repo up-to-date is probably for the best 🙂

PR #114, opened 2019-10-19
Allow users not to have a `schema` directory 

jamiebrynes7:
In `cargo-spatial` we assumed that there would be a `schema` directory next to the `Spatial.toml`. This may not be true, as is the case with our `test-suite` crate.

This PR removes the default assumption of this `schema` directory and removes some unused/low-impact dependencies for the `cargo-spatial` tool.

### DIFF
--- a/cargo-spatial/src/codegen.rs
+++ b/cargo-spatial/src/codegen.rs
@@ -1,12 +1,11 @@
 use crate::config::Config;
+use crate::format_arg;
 use log::*;
 use spatialos_sdk_code_generator::{generator, schema_bundle};
-use std::ffi::OsString;
 use std::fs::{self, File};
 use std::io::prelude::*;
 use std::path::*;
 use std::process::Command;
-use tap::*;
 
 /// Performs code generation for the project described by `config`.
 ///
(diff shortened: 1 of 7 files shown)


cargo-spatial/src/codegen.rs
```diff
@@ -92,22 +89,3 @@ pub fn run_codegen(config: &Config) -> Result<(), Box<dyn std::error::Error>> {
...
-/// some idiosyncrasies with schema_compiler and protoc. Currently,
-/// schema_compiler and protoc get tripped up if you have paths with mixed path
-/// separators (i.e. mixing '\' and '/'). This function normalizes paths to use
-/// the same separators everywhere, ensuring that we can be robust regardless of
-/// how the user specifies their paths. It also removes any current dir segments
-/// ("./"), as they can trip up schema_compiler and protoc as well.
-///
-/// Improbable has noted that they are aware of these issues and will fix them
-/// at some point in the future.
-fn normalize<P: AsRef<std::path::Path>>(path: P) -> PathBuf {
```
randomPoison: Glad to see this is no longer needed 🎉

Issue #67, opened 2019-02-08
Add Ergonomic API for interacting with component updates


jamiebrynes7:
Constructing component updates is pretty verbose right now: 

I.e. - 

```
improbable::PositionUpdate {
                    coords: Some(improbable::Coordinates {
                        x: angle.sin() * 10.0,
                        y: 0.0,
                        z: angle.cos() * 10.0,
                    }),
                },
```

vs 

```
improbable::PositionUpdate::new()
	.with_coords(improbable::Coordinates { 
		x: angle.sin() * 10.0, 
		y: 0.0, 
		z: angle.cos() * 10.0,
	})
```

This problem will get worse if you have a component with many fields. 

As a sidenote - constructors for generated types would help as well :)

PR #115, opened 2019-10-20
Initial events implementation

jamiebrynes7:
This PR implements a first pass at events. This roughly follows the discussion that we had on Discord a little while ago. 

---

A couple discussion points: 

#### 1. Events struct representation

Given schema similar to: 
```
type PayloadA {}
type PayloadB {}

component Foo {
	id = 1;
	event PayloadA firstEvent;
	event PayloadB secondEvent;
}
```
The events struct currently looks like:
```rust
pub struct ComponentUpdateData<C> {
	pub fields: C::Update,
	pub events: C::Events
}

...

pub struct FooEvents {
	pub firstEvent: Vec<PayloadA>,
	pub secondEvent: Vec<PayloadB>
}
```
Whereas it could look like:
```rust
pub struct ComponentUpdateData<C> {
	pub fields: C::Update,
	pub events: Vec<C::Events>
}

pub enum FooEvent {
	FirstEvent(PayloadA),
	SecondEvent(PayloadB)
}
```
The reason I went for the former at not the latter is:
- It was the quickest to implement (the `TypeConversion` trait could be used similarly to updates).
- It doesn't imply some sort of ordering for the events in a given update.

That being said, I don't have a strong preference either way. Opinions very welcome.

##### 2. Generating an `Events` struct for _all_ components. 

We currently create an `Events` struct per component, event if there are no events in that component. We could use the unit type, `()`, to represent the events in the `Component` trait implementation instead. This does come at the cost of adding a lot of conditional logic into the code generator. 

Its worth noting we do the same for the commands currently, although these have less of a side effect. With how its currently implemented, sending a Position component update requires me to instantiate a `improbable::PositionEvents` struct, which feels... odd.

##### 3. Naming

The hardest decisions of them all. I stuck with `ComponentUpdateData` since that's what we originally discussed, but now we have:

- `trait ComponentUpdate` (update fields only)
- `trait ComponentData` (at rest component fields)
- `struct ComponentUpdateData`	(update fields and events)

Which could get somewhat confusing. Suggestions welcome to find some clearer names. 

##### 4. Verbose API

This change introduces a little more boilerplate into creating and sending updates, it may be worth visiting #67 soon to compensate. One small thing we could do is add `#[derive(Default)]` to `ComponentUpdateData` to allow users to skip the fields or events when constructing updates.

See https://github.com/jamiebrynes7/spatialos-sdk-rs/pull/116 for an example of how we could alleviate this.

### DIFF
--- a/project-example/schema/example.schema
+++ b/project-example/schema/example.schema
@@ -62,5 +62,6 @@ type Payload {
 component EventTest {
     id = 2003;
 
+    int32 field = 1;
     event Payload payload;
 }
--- a/project-example/schema/example.schema
+++ b/project-example/schema/example.schema
@@ -54,3 +54,13 @@ component EnumTestComponent {
 
     TestEnum test = 1;
 }
+
+type Payload {
+    int32 data = 1;
+}
+
+component EventTest {
+    id = 2003;
+
+    event Payload payload;
+}
--- a/project-example/src/generated.rs
+++ b/project-example/src/generated.rs
@@ -512,6 +512,9 @@ impl EventTestUpdate {
 impl ComponentUpdate<EventTest> for EventTestUpdate {
     fn merge(&mut self, update: EventTestUpdate) {
         if update.field.is_some() { self.field = update.field; }
+        let mut payload = update.payload;
+        self.payload.append(&mut payload);
+
     }
 }
 
--- a/project-example/src/generated.rs
+++ b/project-example/src/generated.rs
@@ -160,49 +160,53 @@ impl ComponentData<EntityIdTest> for EntityIdTest {
     fn merge(&mut self, update: EntityIdTestUpdate) {
         if let Some(value) = update.eid { self.eid = value; }
     }
+
+    fn merge_ref(&mut self, update: &EntityIdTestUpdate) {
+        let copy = EntityIdTestUpdate {
+            eid: update.eid, 
+        };
+
+        self.merge(copy);
+    }
 }
 
 #[derive(Debug, Clone, Default)]
 pub struct EntityIdTestUpdate {
     pub eid: Option<spatialos_sdk::worker::EntityId>,
 }
-impl TypeConversion for EntityIdTestUpdate {
-    fn from_type(input: &SchemaObject) -> Result<Self, String> {
+
+impl EntityIdTestUpdate {
+    pub fn from_schema(update: &SchemaComponentUpdate) -> Result<Self, String> {
+        let fields = update.fields();
+        let events = update.events();
+
         let mut output = Self {
-            eid: None,
-        };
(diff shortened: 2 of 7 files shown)


dgavedissian:
Will review this properly later, but in the other worker SDKs, it basically has option (1), but instead the `ComponentUpdateData` is rolled into `ComponentUpdate`, so given the schema:

```
type PayloadA {}
type PayloadB {}

component Foo {
	id = 1;
        int32 field = 1;
        float other_field = 2;
	event PayloadA first_event;
	event PayloadB second_event;
}
```

this would generate something like:
```
type FooUpdate {
  field: i32,
  other_field: f32,
  first_event: Vec<PayloadA>,
  second_event: Vec<PayloadB>,
}
```

With this approach, we don't need to introduce a new `ComponentUpdateData` type (and keep using `C::Update`), and it also solves the 2nd problem about generating event objects for things which don't have events. In that case, the update just has no event fields.

jamiebrynes7:
@dgavedissian In which case, I'd hold on off on a full review then :wink:

jamiebrynes7:
Although that does make the `merge` method on the `ComponentData` trait a little trickier since it consumes the `C::Update` so you'd lose access to the attached events

https://github.com/jamiebrynes7/spatialos-sdk-rs/blob/master/spatialos-sdk/src/worker/component.rs#L17

dgavedissian:
What's stopping the user just moving the events out of the update before merging it?

dgavedissian:
Ehh, I guess because the update is not mutable. I forgot that this is not C++.

jamiebrynes7:
Actually given how we currently have things implemented, you must `.clone()` an update before you can merge it anyway since we only give out references to the update: 

https://github.com/jamiebrynes7/spatialos-sdk-rs/blob/master/spatialos-sdk/src/worker/op.rs#L583

Although that also makes me concerned since I'd rather not _need_ to clone multiple `Vec` to merge my data. We could have the implementer of `merge` clone the required data instead?

dgavedissian:
That said, I do think it's OK to just drop the events if the user chooses to merge a component update into an at-rest data type, cause events make no sense in that context anyway.

I think it is probably best to implement two functions, one that takes a `C::Update` and another that takes a `&C::Update` (and make the latter call `.clone()` then call the former). That way, a user is not always locked out of an optimisation in any case.

jamiebrynes7:
Discussed offline: 

@dgavedissian's suggestion to: 

> I think it is probably best to implement two functions, one that takes a C::Update and another that takes a &C::Update (and make the latter call .clone() then call the former). That way, a user is not always locked out of an optimisation in any case.

Seems like the best way out. We will add another trait function `merge_ref` which will internally clone the fields (but not the events) and call `merge`.

PR #118, opened 2019-10-29
add default snapshot to project config

johnpmayer:
This fixes the following bug when running `cargo spatial local launch`

```
Encountered an error during command execution.                                 Snapshot file is not set. Please provide a snapshot file using either t
he --snapshot flag or the 'defaultSnapshotUploadFilepath' field under 'snapshot' in the project configuration file
```

Could be that there's an alternative approach, but this felt easy enough.

### DIFF
--- a/project-example/spatialos.json
+++ b/project-example/spatialos.json
@@ -8,5 +8,6 @@
   "clientWorkers": [
     "./rust_client.json"
   ],
-  "launchConfiguration": "world.json"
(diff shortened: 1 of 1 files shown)


jamiebrynes7:
Duplicated your commit over to https://github.com/jamiebrynes7/spatialos-sdk-rs/pull/119 so the CI will run (it won't from forks since there are secrets 😢)

PR #119, opened 2019-10-29
add default snapshot to project config

jamiebrynes7:
> Duped from #118 such that CI can run 

This fixes the following bug when running `cargo spatial local launch`

```
Encountered an error during command execution.                                 Snapshot file is not set. Please provide a snapshot file using either t
he --snapshot flag or the 'defaultSnapshotUploadFilepath' field under 'snapshot' in the project configuration file
```

Could be that there's an alternative approach, but this felt easy enough.

### DIFF
--- a/.github/workflows/premerge.yml
+++ b/.github/workflows/premerge.yml
@@ -44,11 +44,13 @@ jobs:
 
       - name: Generate project-example code
         run: pushd project-example && cargo spatial --verbose codegen && popd
+        shell: bash
         env:
           SPATIAL_LIB_DIR: "../dependencies"
 
       - name: Generate test-suite code
         run: pushd test-suite && cargo spatial --verbose codegen && popd
+        shell: bash
         env:
           SPATIAL_LIB_DIR: "../dependencies"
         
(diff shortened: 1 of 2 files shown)


jamiebrynes7:
Huh. Turns out the default shell changed for Windows machines since the last PR went through 😰 

Check on #114: 

```
Run pushd project-example && cargo spatial --verbose codegen && popd
  pushd project-example && cargo spatial --verbose codegen && popd
  shell: C:\windows\system32\cmd.exe /D /E:ON /V:OFF /S /C "CALL "{0}""
  env:
    SPATIAL_LIB_DIR: ../dependencies
```

Check on this one: 

```
Run pushd project-example && cargo spatial --verbose codegen && popd
  pushd project-example && cargo spatial --verbose codegen && popd
  shell: C:\Program Files\PowerShell\6\pwsh.EXE -command ". '{0}'"
  env:
    SPATIAL_LIB_DIR: ../dependencies
```

Should be a simple fix though - I'll get that done and merged later today @johnpmayer 😄

Issue #121, opened 2019-10-31
Schema serialization rewrite and migration


randomPoison:
After a couple of fruitful-but-failed attempts at rewriting the schema serialization logic (#89, #101), I've now got a better idea of how to improve the serialization system. At this point, the problem is figuring out how to migrate to the improved system without having to do a single, massive PR. The goal of this ticket is to lay out a plan for that migration, and to act as a tracker for the progress.

The previous pass at rewriting the serialization system is archived in a [fork of the repository](https://github.com/randomPoison/spatialos-sdk-rs/tree/serialization-rewrite).

[Progress is also tracked in a project](https://github.com/jamiebrynes7/spatialos-sdk-rs/projects/5).

## Rationale

There are a number of problems that I am attempting to solve with this rewrite:

* **Safety issues with schema data objects**. The way that the schema data objects are exposed is wildly unsafe. It does not enforce the required ownership for data, exposing soundness holes in safe Rust code.
* **Better encode schema serialization rules** using Rust's type system. The current setup doesn't do a good job statically enforcing that schema types match the semantics of the schema language.
* **Provide a more idiomatic API**. If possible, we should more closely mirror how [serde](https://serde.rs/) handles serialization.

## Migration Steps

The goal is to break this work up into smaller steps that can each be merged in one at a time in the hopes of making the migration work more manageable.

* [x] Restructure module hierarchy. This is an easy first step and will help make future refactorings less messy. (#122)
  * [x] Move `schema` out of `internal`.
  * [x] Move schema primitive types into submodule of `schema`.
  * [x] Move schema object types into separate submodules of `schema`.
* [x] Remove `SchemaFieldContainer`. The current setup is incompatible with the safety fixes, so getting this done will help keep the later changes smaller. (#124)
* [x] Address schema object safety issues. This is relatively self-contained work, and can be done without changing how serialization itself is handled.
  * [x] `SchemaObject` (#127)
  * [x] `SchemaComponentData` (#128)
  * [x] `SchemaComponentUpdate` (#129)
  * [x] `SchemaCommandRequest` (#129)
  * [x] `SchemaCommandResponse` (#129)
* [x] Rework schema serialization setup.
  * [x] Compare the current setup to the ideal setup.
  * [x] Determine more granular migration steps.
[...]

PR #127, opened 2019-11-11
Fix soundness holes in SchemaObject

randomPoison:
> Part of #121 

This is the first of the schema data object soundness fixes. It follows the pattern described in #101 to make the schema data types sound. In summary:

* Make `SchemaObject` an opaque type that is never actually instantiated.
* Cast and dereference a `*mut Schema_Object` to either a `&SchemaObject` or a `&mut SchemaObject`, depending on what is valid for the context.
* Internally within the SDK, convert a `&SchemaObject` or `&mut SchemaObject` back to a `*mut Schema_Object` to pass into the C API.

This provides an idiomatic Rust API that allows us to correctly track lifetimes and enforce mutability constraints.

---

For the most part, this required very little change to the internals of the SDK or public API, and no changes to code generation. We were already passing `&SchemaObject` to any functions, so the only public facing changes were to change functions that returned a `SchemaObject` to instead return either a `&SchemaObject` or `&mut SchemaObject`.

Internally, the only changes were to how we construct `SchemaObject` instances and how we access the raw pointer from them.

Should be fairly straightforward overall :grin:

### DIFF
--- a/spatialos-sdk/Cargo.toml
+++ b/spatialos-sdk/Cargo.toml
@@ -13,6 +13,7 @@ spatialos-sdk-sys = { path = "../spatialos-sdk-sys"}
 futures = "0.1"
 inventory = "0.1"
 lazy_static = "1.3"
+static_assertions = "1.1.0"
 
 [dev-dependencies]
 structopt = "0.3"
--- a/spatialos-sdk/Cargo.toml
+++ b/spatialos-sdk/Cargo.toml
@@ -13,8 +13,8 @@ spatialos-sdk-sys = { path = "../spatialos-sdk-sys"}
 futures = "0.1"
 inventory = "0.1"
 lazy_static = "1.3"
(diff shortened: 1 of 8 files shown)


spatialos-sdk/Cargo.toml
```diff
@@ -13,6 +13,7 @@ spatialos-sdk-sys = { path = "../spatialos-sdk-sys"}
 futures = "0.1"
 inventory = "0.1"
 lazy_static = "1.3"
+static_assertions = "1.1.0"
```
jamiebrynes7: Personally prefer marking this as `dev-dependency` and wrapping the assertions with `#[cfg(test)]` so as to eliminate a dependency for an end user.

Since we run `cargo test` in CI, we will always check these assertions as a pre-merge check

randomPoison: Ah, good idea! 😁

PR #128, opened 2019-11-12
SchemaComponentData soundness

randomPoison:
> Part of #121 

Further work re-implementing the work from #101, this time patching up the soundness holes in `SchemaComponentData`. This is very similar to the changes in #127, with the addition step of adding the `Owned` smart pointer type.

`Owned` is like `Box`, but used only for the schema data objects. This abstraction is necessary because even when we "own" a `Schema_ComponentData`, we only have a pointer and not the struct itself. As such, we can't represent it in Rust as a struct on its own or the ownership interactions will (to use a technical term) get all fucky. In this case, we need to create our own smart pointer type instead of using `Box` directly since each schema data object objects has its own deallocation function in the C API.

As a helper for `Owned`, we also define the `Ownable` and `OwnableImpl` traits, with `Ownable` just being a public-facing facade that we can use as a trait bound in the public API for `Owned`. The `OwnableImpl` trait contains the real logic, i.e. an associated type for the raw pointer representation of the schema data object and a function for freeing the object when dropped.

I've also removed the `ComponentData` helper type from the `component::internal` module and replaced it with the `ComponentDataRef` type. In doing so, I've made a couple of tweaks to how incoming schema data (from worker ops) is handled:

* Use an `Option<NonNull>` to represent the raw user handle. This enforces stricter null checking and allows us to do a bit less raw pointer manipulation.
* Return a `Cow` when attempting to get the deserialized component data, automatically deserializing the schema object if the user handle is null (i.e. if vtable serialization is disabled for that type).

This is the pattern I took for #101, and I'll continue to convert the remaining schema objects types this way if it looks good to yall 😁

### DIFF
--- a/spatialos-sdk/src/worker/schema/component_data.rs
+++ b/spatialos-sdk/src/worker/schema/component_data.rs
@@ -31,10 +31,6 @@ impl SchemaComponentData {
         &*(raw as *mut _)
     }
 
-    pub(crate) unsafe fn from_raw_mut<'a>(raw: *mut Schema_ComponentData) -> &'a mut Self {
-        &mut *(raw as *mut _)
-    }
-
     pub(crate) fn as_ptr(&self) -> *mut Schema_ComponentData {
         self as *const _ as *mut _
     }
--- a/spatialos-sdk/src/worker/schema/component_data.rs
+++ b/spatialos-sdk/src/worker/schema/component_data.rs
@@ -55,5 +55,11 @@ impl OwnableImpl for SchemaComponentData {
     }
 }
 
+// SAFETY: It should be safe to send a `SchemaComponentData` between threads, so long as
+// it's only ever accessed from one thread at a time. It has unsynchronized internal
+// mutability (when getting an object field, it will automatically add a new object
+// if one doesn't already exist), so it cannot be `Sync`.
+unsafe impl Send for SchemaComponentData {}
+
 assert_impl_all!(SchemaComponentData: Send);
 assert_not_impl_any!(SchemaComponentData: Sync);
--- a/spatialos-sdk/src/worker/schema/component_data.rs
+++ b/spatialos-sdk/src/worker/schema/component_data.rs
@@ -1,29 +1,59 @@
-use crate::worker::schema::SchemaObject;
+use crate::worker::schema::{owned::*, SchemaObject};
 use spatialos_sdk_sys::worker::*;
+use static_assertions::*;
+use std::marker::PhantomData;
 
+/// Serialized schema data for a component, owned by the Rust SDK.
+///
+/// For maximum efficiency, the serialized data may borrow data from the component
+/// used to create an `OwnedComponentData` instance. The lifetime parameter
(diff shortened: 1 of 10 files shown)


spatialos-sdk/src/worker/schema/component_data.rs
```diff
@@ -1,29 +1,66 @@
...
+    }
+
+    // Methods for raw pointer conversion.
+    // -----------------------------------
+
+    pub(crate) unsafe fn from_raw<'a>(raw: *mut Schema_ComponentData) -> &'a Self {
+        &*(raw as *mut _)
+    }
+
+    pub(crate) fn as_ptr(&self) -> *mut Schema_ComponentData {
```
jamiebrynes7: Shouldn't this take `&mut self`?

jamiebrynes7: Not that it _really_ matters since you can copy the pointer, but it at least makes the intent explicit

jamiebrynes7:
I really like the `Owned<T>` pattern for this sort of thing. Makes it much more explicit. 

Is there ever a use case where we would mix owned and unowned data? 

If so, would it make sense to have a `MaybeOwned<T>` type such that:

```rust
pub enum MaybeOwned<'a, T: 'a> where T : Ownable
{
	Owned(NonNull<T::Raw>),
	Borrowed(&'a T::Raw)
}
```

such that we can treat owned and unowned data much the same?

I.e. - say I was mixing owned `SchemaComponentData` and unowned one: 

```rust
let mut data = Vec::new(); 
data.push(SchemaComponentData::from_raw(unowned_ptr));
data.push(SchemaComponentData::new());
```

This would compile just fine since data would be of type `Vec<MaybeOwned<'a, SchemaComponentData>>`

spatialos-sdk/src/worker/schema/component_data.rs
randomPoison: There are cases where we want to use a `&SchemaComponentData` with the C API, e.g. in the `fields` method. In an earlier iteration I had two methods for getting the raw poiner:

```
fn as_ptr(&self) -> *mut Schema_ComponentData { ... }

fn as_ptr_mut(&mut self) -> *mut Schema_ComponentData { ... }
```

Both have to return a `*mut` since that's what all of the C API functions take (rightfully so, since most of the methods perform some kind of internal mutation in certain cases), so we can't even use the difference to enforce that you need a `&mut` to get a `*mut`. I think it's ultimately better to just have one way to get the raw pointer out, and use other methods for documenting intent and enforcing safety (e.g. expose all meaningful operations as methods on `SchemaComponentData`, and avoid going through the raw C API outside of this module).

randomPoison:
> Is there ever a use case where we would mix owned and unowned data? If so, would it make sense to have a `MaybeOwned<T>` type

Good question! So far I haven't run into any cases where it makes sense to mix owned and borrowed schema data objects like that, but if we did it might make sense to have something like `MaybeOwned`. We should keep an eye out for any use cases that would benefit from this, but I'd hold off on adding it to the SDK until we have a concrete use case 🙂

PR #130, opened 2019-11-17
Unify handling of data pointer types

randomPoison:
> Part of #121

This PR is a cleanup pass after the previous work done to fix the soundness holes in the schema data object types (#127, #128, and #129).

I've added the `DataPointer` trait to unify the functionality for handling the schema data objects. It provides default functions for converting between Rust references and the raw pointer representation, removing the need to have that logic duplicated for each type. I've also added documentation around the invariants that must be maintained in order to for the implementation to be safe.

Per @jamiebrynes7's suggestion, I've changed `as_ptr` to return a `*const Self::Raw` and added a separate `as_ptr_mut` method. In the cases where a method taking `&self` calls a C function that needs `*mut T`, we manually cast the pointer types. The safety invariants around this are documented in the "Interior Mutability and Thread Safey" section of the `ptr` module docs.

Lastly, I've added a `pointer_type_tests!` macro to generate a set of assertion tests for each of these types to help ensure we're enforcing the necessary safety invariants.

### DIFF
--- a/spatialos-sdk/src/worker/schema/ptr.rs
+++ b/spatialos-sdk/src/worker/schema/ptr.rs
@@ -1,0 +1,101 @@
+//! Handling for schema data types that act like pointers.
+//!
+//! `Schema_Object` and the related schema data "wrapper types" are all represented
+//! as opaque types hidden behind pointers in the C API; You always work with a
+//! pointer to the struct, never an instance of the struct itself. To represent
+//! these in Rust, we define dummy types that can similarly always be put behind a
+//! reference without ever allowing the user to directly "hold" an instance of the
+//! type. This allows us to simply cast the raw pointer to a reference to our dummy
+//! type and preserve the desired semantics of working with a reference, including
+//! the ownership rules and lifetimes that come with it.
+//!
+//! The `DataPointer` trait is implemented for all such types, and provides the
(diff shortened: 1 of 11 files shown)


spatialos-sdk/src/worker/schema/ptr.rs
```diff
@@ -0,0 +1,101 @@
...
+//! reference, is a form of [interior mutability] and is still safe as long as we
+//! maintain the necessary invariants. In particular, our Rust types cannot be
+//! `Sync` because the mutability happening within the C SDK is unsynchronized.
+//! The other key invariant is that we must ensure that a reference returned by a
+//! method taking `&self` can never be invalidated by another method taking `&self`,
+//! however the C API already guarantees this behavior in all cases and so is not
+//! something we specifically have to enforce within the Rust SDK.
+//!
+//! [zst]:  https://doc.rust-lang.org/nomicon/exotic-sizes.html#zero-sized-types-zsts
+//! [rfc]: https://github.com/rust-lang/rfcs/blob/master/text/1861-extern-types.md
```
jamiebrynes7: Didn't know this was a thing! Very neat :)

randomPoison: Me neither! I just happened to come across it while doing research for the documentation :grin:

PR #131, opened 2019-11-17
Unify serialization logic

randomPoison:
> Part of #121 

This PR unifies how schema data serialization is handled, specifically allowing collection types and user-defined types (struct types and enums) to be serialized via the same mechanism as primitive types. This allows us to drastically simplify the serialization code generated for schema-defined types.

* Rename `SchemaPrimitiveField` to `Field` now that it is used for all field types and not just primitives.
* Remove `component::TypeConversion` trait and replace it with `schema::ObjectField` trait. This trait functions largely the same as before, but function names have been changed and functions no longer return `Result`s.
* Add a blanket impl of `Field` for all types implementing `ObjectField`. This ensures we only have to generate code for the simpler `ObjectField` trait, and can serialize object fields in the same manner as primitive fields.
* Add `EnumField` trait and `impl_field_for_enum_field!` helper macro to simulate a blanket impl of `Field` for types implementing `EnumField`. Rust's coherence rules prevent us from directly providing such a blanket impl, as noted in the comments.
* Use `TryFrom` to avoid panicking if deserializing an enum from an unknown discriminant value.
* Add `schema::collections` module with marker types `Optional<T>`, `List<T>`, and `Map<K, V>`. These marker types implement `Field` so that schema collection types can be serialized in the same manner as primitive fields.
* Update code generation accordingly. It's much simpler now!

The most complex part of the code generation at this point is handling component updates, since the generated code still has to check to see if the field is present in the update or not. This also means that collection types aren't handling being cleared correctly. This will be fixed in the next PR.

### DIFF
--- a/spatialos-sdk-code-generator/src/generator.rs
+++ b/spatialos-sdk-code-generator/src/generator.rs
@@ -94,18 +94,18 @@ impl Package {
             }
 
             FieldDefinition_FieldType::Option { inner_type } => {
-                format!("schema::Option<{}>", self.schema_type_name(inner_type)).into()
+                format!("Optional<{}>", self.schema_type_name(inner_type)).into()
             }
 
             FieldDefinition_FieldType::List { inner_type } => {
-                format!("schema::List<{}>", self.schema_type_name(inner_type)).into()
+                format!("List<{}>", self.schema_type_name(inner_type)).into()
             }
 
             FieldDefinition_FieldType::Map {
                 key_type,
                 value_type,
             } => format!(
-                "schema::Map<{}, {}>",
+                "Map<{}, {}>",
                 self.schema_type_name(key_type),
                 self.schema_type_name(value_type),
             )
@@ -211,23 +211,6 @@ impl Package {
         }
     }
 
-    // Some types need to be borrowed when serializing (such as strings or objects). This helper function returns true
-    // if this is required.
-    fn type_needs_borrow(&self, type_ref: &TypeReference) -> bool {
-        match type_ref {
-            TypeReference::Primitive(ref primitive) => match primitive {
-                PrimitiveType::String => true,
-                PrimitiveType::Bytes => true,
-                PrimitiveType::Entity => true,
-                _ => false,
-            },
-            // Enums are always data.
-            TypeReference::Enum(_) => false,
-            // Types always need borrowing, as they're structs.
-            TypeReference::Type(_) => true,
-        }
-    }
-
     // Generates an expression which serializes a field from an expression into a schema object. The generated
     // expression should always have type ().
     fn serialize_field(
@@ -236,125 +219,23 @@ impl Package {
         expression: &str,
         schema_object: &str,
     ) -> String {
-        match field.field_type {
-            FieldDefinition_FieldType::Singular { ref type_reference } => {
-                self.serialize_type(field.field_id, type_reference, expression, schema_object)
-            }
-
-            FieldDefinition_FieldType::Option { ref inner_type } => {
-                let ref_decorator = if self.type_needs_borrow(inner_type) {
-                    "ref "
-                } else {
-                    ""
-                };
-                format!(
-                    "if let Some({}data) = {} {{ {}; }}",
-                    ref_decorator,
-                    expression,
-                    self.serialize_type(field.field_id, inner_type, "data", schema_object)
-                )
-            }
-
-            FieldDefinition_FieldType::List { ref inner_type } => {
-                // If we have a list of primitives, we can just pass a slice directly to add_list.
-                match inner_type {
-                    TypeReference::Primitive(ref primitive) => format!(
-                        "{}.add_list::<{}>({}, &{}[..])",
-                        schema_object,
-                        get_rust_primitive_type_tag(primitive),
-                        field.field_id,
-                        expression
-                    ),
-                    _ => {
-                        let add_item = self.serialize_type(
-                            field.field_id,
-                            inner_type,
-                            "element",
-                            schema_object,
-                        );
-                        format!("for element in ({}).iter() {{ {}; }}", expression, add_item)
-                    }
-                }
-            }
-
-            FieldDefinition_FieldType::Map {
-                ref key_type,
-                ref value_type,
-            } => format!(
-                "{}.add::<Map<{}, {}>>({}, {})",
-                schema_object,
-                self.schema_type_name(key_type),
-                self.schema_type_name(value_type),
-                field.field_id,
-                expression,
-            ),
-        }
-    }
-
-    // Generates an expression which serializes a value from an expression into a schema object. The generated
-    // expression should always have type ().
-    fn serialize_type(
-        &self,
-        field_id: u32,
-        value_type: &TypeReference,
-        expression: &str,
-        schema_object: &str,
-    ) -> String {
-        match value_type {
-            TypeReference::Primitive(ref primitive) => format!(
-                "{}.add::<{}>({}, {})",
-                schema_object,
-                get_rust_primitive_type_tag(primitive),
-                field_id,
-                expression,
-            ),
-            TypeReference::Enum(_) => format!(
-                "{}.add::<SchemaEnum>({}, &{}.as_u32())",
-                schema_object, field_id, expression
-            ),
-            TypeReference::Type(ref type_ref) => {
-                let type_definition =
-                    self.rust_fqname(&self.get_type_definition(type_ref).qualified_name);
-                format!(
-                    "<{} as TypeConversion>::to_type(&{}, &mut {}.add_object({}))?",
-                    type_definition, expression, schema_object, field_id
-                )
-            }
-        }
+        format!(
+            "{}.add::<{}>({}, {})",
+            schema_object,
+            self.field_type_name(&field.field_type),
+            field.field_id,
+            expression,
+        )
     }
 
     // Generates an expression which deserializes a field from a schema field 'schema_field'.
     fn deserialize_field(&self, field: &FieldDefinition, schema_field: &str) -> String {
-        match field.field_type {
-            FieldDefinition_FieldType::Singular { ref type_reference } => {
-                self.deserialize_type(field.field_id, type_reference, schema_field)
-            }
-
-            FieldDefinition_FieldType::Option { ref inner_type } => {
-                let count_expr = self.count_field(field.field_id, inner_type, schema_field);
-                let deserialize_expr =
-                    self.deserialize_type(field.field_id, inner_type, schema_field);
-                format!(
-                    "if {} > 0 {{ Some({}) }} else {{ None }}",
-                    count_expr, deserialize_expr
-                )
-            }
-
-            FieldDefinition_FieldType::List { ref inner_type } => {
-                self.deserialize_list(field.field_id, inner_type, schema_field)
-            }
-
-            FieldDefinition_FieldType::Map {
-                ref key_type,
-                ref value_type,
-            } => format!(
-                "{}.get::<Map<{}, {}>>({})",
-                schema_field,
-                self.schema_type_name(key_type),
-                self.schema_type_name(value_type),
-                field.field_id
-            ),
-        }
+        format!(
+            "{}.get::<{}>({})",
+            schema_field,
+            self.field_type_name(&field.field_type),
+            field.field_id
+        )
     }
 
     fn deserialize_update_field(&self, field: &FieldDefinition, schema_field: &str) -> String {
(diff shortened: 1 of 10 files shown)


spatialos-sdk/src/worker/schema.rs
```diff
@@ -41,3 +109,151 @@ pub trait SchemaPrimitiveField {
...
+/// [`Field`]: trait.Field.html
+pub trait ObjectField: Sized + Clone {
+    /// Deserializes an instance of this type from a [`SchemaObject`].
+    ///
+    /// Note that deserialization can never fail as the underlying SpatialOS API ensures
+    /// that missing or invalid fields will always provide a valid default value during
+    /// deserialization. As such, if the data in `object` doesn't match what is expected
+    /// by the deserialization logic (e.g. due to a version mismatch between the current
+    /// schema version and the version `object` was generated from), the returned object
+    /// will have default values for any fields that are missing in `object` or have a
```
jamiebrynes7: Rather than default values, it could be garbage data I think

spatialos-sdk/src/worker/schema.rs
```diff
@@ -41,3 +109,151 @@ pub trait SchemaPrimitiveField {
...
+    }
+
+    fn add(object: &mut SchemaObject, field: FieldId, value: &Self::RustType) {
+        value.into_object(object.add_object(field));
+    }
+}
+
+/// The error type returned when a conversion to a schema enum type fails.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct UnknownDiscriminantError;
```
jamiebrynes7: Should this have a field specifying what the schema value was when it failed? Could aid in debugging if this error ever arose

I.e. - 

```rust
pub struct UnknownDiscriminantError(u32);
```

spatialos-sdk/src/worker/schema.rs
```diff
@@ -41,3 +109,151 @@ pub trait SchemaPrimitiveField {
...
+
+/// A type that corresponds to a [schemalang enumeration][schema-enum].
+///
+/// Schemalang enums are C-like enums (i.e. they only have a discriminant and do not
+/// contain any additional data) represented as `u32` values when serialized. They
+/// default to the first variant defined in their schema declaration.
```
jamiebrynes7: What do you mean by "default to the first variant"? Who is defaulting what here?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -41,3 +109,151 @@ pub trait SchemaPrimitiveField {
...
+    ($type:ty) => {
+        impl $crate::worker::schema::Field for $type {
+            type RustType = Self;
+
+            fn get_or_default(
+                object: &$crate::worker::schema::SchemaObject,
+                field: $crate::worker::schema::FieldId,
+            ) -> Self::RustType {
+                Self::try_from(object.get::<$crate::worker::schema::SchemaEnum>(field))
+                    .unwrap_or_default()
```
jamiebrynes7: It would be nice to be able to not lose the fact that there was an error here.

spatialos-sdk/src/worker/schema/collections.rs
```diff
@@ -0,0 +1,156 @@
...
+            object.add::<T>(field, value);
+        }
+    }
+
+    fn index(_object: &SchemaObject, _field: FieldId, _index: usize) -> Self::RustType {
+        panic!("Optional fields cannot be indexed into")
+    }
+
+    fn count(_object: &SchemaObject, _field: FieldId) -> usize {
+        panic!("Optional fields cannot be counted")
```
jamiebrynes7: Feels weird that we panic here since optional values do have a count (either 0 or 1). Is this to indicate that this function shouldn't ever be called during serialization? 

If so, maybe this should be `unreachable!()`? Although that also feels a little wonky

spatialos-sdk/src/worker/component.rs
```diff
@@ -139,23 +130,22 @@ impl<'a> ComponentDataRef<'a> {
         }
     }
 
-    // NOTE: We manually declare that the component impl `TypeConversion` and `Clone`
+    // NOTE: We manually declare that the component impl `ObjectField`
     // here, but in practice this will always be true for all component types. Future
     // iterations should clean this up such that the `Component` trait can imply these
     // other bounds automatically (i.e. by making them super traits of `Component`).
```
jamiebrynes7: Is it worth doing this note in this PR?

spatialos-sdk-code-generator/src/generator.rs
```diff
@@ -1,10 +1,11 @@
 use crate::schema_bundle::*;
 use heck::CamelCase;
+use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::{BTreeMap, BTreeSet};
 use std::rc::Rc;
 
-fn get_rust_primitive_type_tag(primitive_type: &PrimitiveType) -> &str {
+fn get_rust_primitive_type_tag(primitive_type: &PrimitiveType) -> &'static str {
```
jamiebrynes7: Nit - I know you didn't add this but the name `get_rust_primitive_type_tag` is somewhat confusing

spatialos-sdk-code-generator/src/generator.rs
```diff
@@ -201,141 +219,23 @@ impl Package {
...
-                )
-            }
-        }
+        format!(
+            "{}.add::<{}>({}, {})",
+            schema_object,
+            self.field_type_name(&field.field_type),
+            field.field_id,
+            expression,
+        )
```
jamiebrynes7: This is beautiful 👏

spatialos-sdk/src/worker/schema.rs
randomPoison: Huh, is that documented anywhere? I swear I remember @dgavedissian saying you'd get default values if you asked the schema API for fields that don't exist (or asking for an existing field with the wrong type), but I could be putting words in his mouth 😛

spatialos-sdk/src/worker/schema.rs
randomPoison: I considered that, but in any case where you're calling the `TryFrom`, you'll already have the invalid value; You're returning the unknown discriminant back to the code that gave it to you in the first place. That said, I'm not sure there's much of a downside to including the value in the error anyway, maybe just some missed opportunities for optimization since the error won't be a zero sized type anymore?

I'll add that in, since you're right about it possibly being useful in debugging (i.e. you can log the error object and it'll include the unknown discriminant).

spatialos-sdk/src/worker/schema.rs
randomPoison: I guess the `Default` impl is technically determined by code generation, though I was operating under the assumption that trying to get a non-existent field from a schema object would return a default value of 0, which would line up with the first variant of the enum. As I type this out, I realize that might not actually be a valid assumption 🤔

spatialos-sdk/src/worker/schema/collections.rs
randomPoison: Optional fields can't be counted since you can't have a `list<option<T>>` in the schema. This function will be removed entirely once I split off a separate `IndexedField` trait 🙂

spatialos-sdk/src/worker/schema.rs
randomPoison: So, per my previous comment, I had assumed that schema enums can reasonably default to their first variant, which many not be a valid assumption. Is there any concept of defaults for enums in schemalang? If not, handling this case may actually necessitate that deserialization return a `Result` 😲

spatialos-sdk/src/worker/component.rs
randomPoison: Yeah, that would be a good idea 😁

spatialos-sdk/src/worker/schema/collections.rs
jamiebrynes7: Ah sounds good! :)

spatialos-sdk/src/worker/schema.rs
jamiebrynes7: Sadly schema enums don't always start at 0 :( 

```
enum MyEnum 
{
	VALUE = 1;
}
```

is perfectly valid schema. This also means that getting a value of `1` for a field of type MyEnum maps to `VALUE` and a value of `0` is 🤷‍♂ 

https://github.com/spatialos/gdk-for-unity/issues/960 :wink: 


---

I'm not sure exactly what the behaviour of calling `Schema_GetUint32(obj, 2);` where field 2 doesn't exist. Paging @dgavedissian!

spatialos-sdk/src/worker/schema.rs
jamiebrynes7: > for an existing field with the wrong type

This I'm pretty sure returns garbage data (I think I've ran into this before). But I'd need to double check.

For a non-existent field, I'm really not sure!

spatialos-sdk/src/worker/schema.rs
jamiebrynes7: We finally found the use case where deserialization can fail 🙃 

Its always enums that break things :(

spatialos-sdk/src/worker/schema.rs
randomPoison: Huh. Is there any way to verify that a given field matches the expected type? That could be another potential case for failure on deserialization 🤔

spatialos-sdk/src/worker/schema.rs
randomPoison: Ah, I was just going to ask how this is handled in other GDKs! Seems like the answer is "it isn't" 😭

spatialos-sdk/src/worker/schema.rs
randomPoison: Huh. Is there any way to verify that a given field matches the expected type? That could be another potential case for failure on deserialization 🤔

jamiebrynes7: Even worse we won't know that the deserialization has failed if we have the wrong types!

dgavedissian: Bear in mind that `Schema_GetXX` functions under the hood performs lazy deserialization of serialized protobuf data. Therefore, if you get the type wrong, there's a chance that it'll interpret the bytes in the wrong way (so you get garbage out). However, if you try and read a non-existent field, then it'll just return the default value (always 0 / null).

dgavedissian: We specify in our docs [here](https://docs.improbable.io/reference/14.2/shared/schema/evolution#making-changes-to-schema) that schema is only forwards/backwards compatible if you never reuse field IDs. So, you shouldn't expect it to really happen that component data doesn't match the schema (and the field IDs are different), unless the user has really messed up.

spatialos-sdk/src/worker/schema.rs
dgavedissian: Yeah. Enums aren't really a thing in the `Schema_` serialization layer. If you call `Schema_GetUint32(obj, 2)`, then it'll just return 0 if field ID 2 doesn't exist as Jamie said. I suggest that in codegen you special case the enum field in a way like this:

```
if Schema_GetUint32Count(obj, 2) == 0 {
  MyEnum::default()
} else {
  Schema_GetUint32(obj, 2)
```

randomPoison: That's effectively what the current implementation in the PR is doing: If the value doesn't match a known variant, you get the default value out. I'm wondering if that's the right thing to do, though. The user has no way to specify which variant is the default, and in some cases there might not be a meaningful default. Silently spitting out potentially invalid data seems like the wrong thing to do, at least for a Rust API.

Schemalang [supports custom annotations](https://docs.improbable.io/reference/14.2/shared/schema/reference#annotations), so I suppose what we can do is default to producing an error if the value is unknown, but allow the user to specify a default value to use, which would allow serialization to succeed in the face of an unknown variant.

spatialos-sdk/src/worker/schema.rs
randomPoison: I'd like to do that in a separate PR if you're cool with that. It's going to be a fairly noisy change, and there are some other error cases that I was previously ignoring that I'd like to add handling for at the same time. For now, I can either leave the logic for providing a default value (which may still be reasonable to do, per @dgavedissian's suggestion), or I can have it panic if value is invalid for the enum.

spatialos-sdk/src/worker/schema.rs
randomPoison: What's the "official" way to handle non-compatible schema changes? Based on the lack of error handling in the C API and the things you've said, it seems like it comes down to:

* Detectable errors (missing fields and invalid enums) are handled in an SDK-specific way, with the bulk of the handling being done by generated code.
* Undetectable errors (field is present but types are mismatched) produce garbage data but otherwise will not crash or produce errors.

Ultimately, we'll have to do *something* if the user makes a non-compatible change to their schema, if possible I'd like to keep that behavior as consistent with other language SDKs as possible. Ideally I'd like to return an error for any error case, but if some errors are undetectable then keeping things consistent is the next best option.

spatialos-sdk/src/worker/schema.rs
randomPoison: Okay, I did you one better and also added the type name to the error context, that way you can get actually useful error messages like:

```
Found unknown discriminant value 7 for enum `my_crate::schema::Foo` while deserializing schema data
```

Type names are `&'static str`, so it doesn't require any allocation or nothing :tada:

spatialos-sdk/src/worker/component.rs
randomPoison: Updated `Component` to be a sub-trait of `ObjectField`. That removed the need for `from_data` and `to_data`. I also added some helper methods to `SchemaComponentData` to make serialization easier, and removed the trait bounds noted in this comment.

spatialos-sdk-code-generator/src/generator.rs
randomPoison: Renamed to `primitive_type_name`, hopefully that's a bit better 😅

spatialos-sdk/src/worker/schema.rs
jamiebrynes7: Lets leave the current behaviour and revisit in a future PR 😄

PR #135, opened 2019-11-28
Fix CLI for example project

randomPoison:
I ran into an issue where the executable generated by `cargo spatial local launch` was expecting different command line parameters than what we intended, specifically it was expecting `--WORKER_ID` instead of `--worker-id`. When running the executable with `cargo run` directly, it would have the expected CLI.

After some digging, it looks like the difference was because `cargo install` (which is used by `cargo spatial local launch` to build the worker) ignores the project's lock file by default, causing the version built for the local deployment to be different than the version built normally. To address these issues I've made two changes:

* Remove explicit `structopt(name)` attributes. This was causing structopt's default behavior of renaming all fields to be kebab-case to fail. This is due to a change introduced in [structopt v0.3.4](https://github.com/TeXitoi/structopt/blob/master/CHANGELOG.md#v034-2019-11-08).
* Specify `--locked` flag when building local project. This will ensure `cargo spatial local launch` will use the same dependency versions as `cargo run`, cutting down on possible confusion.

### DIFF
--- a/cargo-spatial/src/local.rs
+++ b/cargo-spatial/src/local.rs
@@ -18,7 +18,7 @@ pub fn launch(config: &Config, launch: &LocalLaunch) -> Result<(), Box<dyn std::
     // Run codegen and such.
     crate::codegen::run_codegen(config)?;
 
-    // Use `cargo install` to build workers and copy the exectuables to the build
+    // Use `cargo install` to build workers and copy the executables to the build
     // directory.
     //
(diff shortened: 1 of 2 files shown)


randomPoison:
Ah, good point! I totally forgot that you were already fixing these 😅 I'd suggest adding the `--locked` flag to `cargo spatial local launch` in your PR, but I'll close this one since it's not needed.

PR #136, opened 2019-11-29
Fallible deserialization

randomPoison:
> Part of #121

Have deserialization functions return a `Result` again (this was previously removed in #131 and is being re-introduced in an improved form). Only deserialization is fallible; Serialization cannot fail in a recoverable way for a well-formed Rust type (hooray for type-safety).

To provide meaningful error information to the user, I've added a `schema::Error` type that captures the specific reason that deserialization failed. It supports chaining multiple errors together in order to create a "backtrace" through deeply-nested data.

While working on this, I also setup `SchemaBool` and `SchemaEntityId` to be generated via `impl_primitive_field!`. This required some tweaks to the macro to use `mem::transmute` since these types don't line up with the C API directly. I've added comments on why this setup is still safe.


spatialos-sdk/src/worker/schema/collections.rs
```diff
@@ -109,13 +115,16 @@ where
...
+    fn get_update(
+        update: &SchemaComponentUpdate,
+        field: FieldId,
+    ) -> Result<Option<Self::RustType>> {
         if update.is_field_cleared(field) {
-            Some(Default::default())
+            Ok(Some(Default::default()))
         } else if T::count(update.fields(), field) > 0 {
-            Some(Self::get_or_default(update.fields(), field))
+            Ok(Some(Self::get(update.fields(), field)?))
```
jamiebrynes7:
```suggestion
            Self::get(update.fields(), field).map(Some)
```

spatialos-sdk/src/worker/schema/collections.rs
```diff
@@ -206,13 +220,16 @@ where
...
+    fn get_update(
+        update: &SchemaComponentUpdate,
+        field: FieldId,
+    ) -> Result<Option<Self::RustType>> {
         if update.is_field_cleared(field) {
-            Some(Default::default())
+            Ok(Some(Default::default()))
         } else if update.fields().object_count(field) > 0 {
-            Some(Self::get_or_default(update.fields(), field))
+            Ok(Some(Self::get(update.fields(), field)?))
```
jamiebrynes7:
```suggestion
            Self::get(update.fields(), field).map(Some)
```

spatialos-sdk/src/worker/schema/collections.rs
```diff
@@ -133,7 +142,7 @@ where
         }
     }
 
-    fn index(_object: &SchemaObject, _field: FieldId, _index: usize) -> Self::RustType {
+    fn index(_object: &SchemaObject, _field: FieldId, _index: usize) -> Result<Self::RustType> {
         panic!("List fields cannot be indexed into")
```
jamiebrynes7: These could now be errors instead of panics?

spatialos-sdk/src/worker/schema.rs
```diff
@@ -103,22 +109,23 @@ pub trait Field {
...
     fn has_update(update: &SchemaComponentUpdate, field: FieldId) -> bool;
 
-    fn get_update(update: &SchemaComponentUpdate, field: FieldId) -> Option<Self::RustType> {
+    fn get_update(
+        update: &SchemaComponentUpdate,
+        field: FieldId,
+    ) -> Result<Option<Self::RustType>> {
         if Self::has_update(update, field) {
-            Some(Self::get_or_default(update.fields(), field))
+            Some(Self::get(update.fields(), field)).transpose()
```
jamiebrynes7:
```suggestion
            Self::get(update.fields(), field)).map(Some)
```

spatialos-sdk/src/worker/component.rs
```diff
@@ -409,9 +418,20 @@ unsafe extern "C" fn vtable_component_update_deserialize<C: Component>(
...
-    1
+    match schema_update.deserialize::<C>() {
+        Ok(deserialized_update) => {
+            *handle_out = handle_allocate(deserialized_update);
+            1
+        }
+
+        // TODO: How should we handle errors occurring during vtable serialization? We could
+        // probably store the whole `Result` in the user handle and pass it to the user when
+        // they try to retrieve the result.
```
jamiebrynes7: Agree with the suggestion here. 😄

spatialos-sdk/src/worker/schema/collections.rs
randomPoison: No, they should still be panics since it's effectively indicative of a bug to call `index` on a collection type. I'll be removing these unnecessary methods in the next PR, too.

spatialos-sdk/src/worker/component.rs
randomPoison: Yeah, the more I think about it the more that seems like the right solution. I want to save that for a separate PR because it's going to require that I make some careful changes to how we handle vtable serialization.

project-example/src/main.rs
```diff
@@ -105,7 +105,7 @@ fn logic_loop(c: &mut WorkerConnection) {
                 // of a new entity before sending any component data for that entity.
                 WorkerOp::AddComponent(add_component) => match add_component.component_id {
                     example::Rotate::ID => {
-                        let rotate = add_component.get::<example::Rotate>().unwrap();
+                        let rotate = add_component.get::<example::Rotate>().unwrap().unwrap();
```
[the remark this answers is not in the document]
randomPoison: I know 😢 Not sure if there's a better solution here, though.

randomPoison:
I applied your suggestions and suppressed the clippy warnings, so you should be good to re-review and approve if it looks good!

spatialos-sdk/src/worker/schema.rs
```diff
@@ -103,22 +109,23 @@ pub trait Field {
...
     fn has_update(update: &SchemaComponentUpdate, field: FieldId) -> bool;
 
-    fn get_update(update: &SchemaComponentUpdate, field: FieldId) -> Option<Self::RustType> {
+    fn get_update(
+        update: &SchemaComponentUpdate,
+        field: FieldId,
+    ) -> Result<Option<Self::RustType>> {
         if Self::has_update(update, field) {
-            Some(Self::get_or_default(update.fields(), field))
+            Self::get(update.fields(), field)).map(Some)
```
jamiebrynes7:
```suggestion
            Self::get(update.fields(), field).map(Some)
```

Sorry I had the syntax slightly wrong in my original suggestion :grimacing:

project-example/src/main.rs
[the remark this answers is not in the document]
jamiebrynes7: Its not ideal, but I imagine higher level abstractions will handle this unwrapping and presenting the data/handling deserialization errors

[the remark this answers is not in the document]
randomPoison: Yeah, that's kinda what I was thinking, e.g. an Amethyst GDK would do most of the work here.

spatialos-sdk/src/worker/schema.rs
randomPoison: Ah yeah, just fixed that :smile:

PR #141, opened 2019-12-31
Refactor vtable module

jamiebrynes7:
When taking a stab at implementing the schema `Entity` type, I found that I would need to add additional "vtable" entries. 

<details><summary>Context for this change</summary>

The schema `Entity` type is serialized as a `SchemaObject` where each component is serialized into a `SchemaObject` field with the field ID equal to that component's ID.

To (de)serialize this properly, we need type erased access to the `ObjectField` implementation for each component (to serialize to/from SchemaObject directly to component).

This would need live in a vtable structure like the one we already have. 
</details>

This PR lays the groundwork for that future change, by refactoring the vtable module and exposing the `VTable` struct instead of the `Worker_ComponentVTable` to the "user" which allows us to add new functions to the vtable.


randomPoison:
Overall the code changes here look good to me 👍 I'm a fan of moving the vtable code into its own module, regardless of other motivations for the change 😁

I'm not familiar with the use cases for serializing an `Entity` as a field of a schema object, so I don't quite understand the context you provided for the change. Is the expectation that user code will manually have to modify the vtable for each type in order to register the necessary serialization function(s)?

jamiebrynes7:
>  Is the expectation that user code will manually have to modify the vtable for each type in order to register the necessary serialization function(s)?

Nope, we will just add new fields to the `VTable` struct and populate them in the `new<C: Component>()` implementation. 

What this will end up looking like is something like:

```rust
pub struct VTable {
	worker_vtable: Worker_ComponentVTable,
	deserialize_component_object: Option<fn(&mut SchemaObject, *mut *mut void)>
	serialize_component_object: Option<fn(&mut SchemaObject, *const void)>
	...
}
```

Those two functions additional functions will be similar to the vtable functions we have:

```rust
unsafe extern "C" fn vtable_component_data_serialize<C: Component>(
    _: u32,
    _: *mut raw::c_void,
    handle: *mut raw::c_void,
    data: *mut *mut Schema_ComponentData,
) {
    let client_data = &*handle.cast::<C>();
    *data = SchemaComponentData::from_component(client_data).into_raw();
}
```

but we are converting from a different data source (a `SchemaObject` rather than a `Schema_ComponentData`). We could use the existing vtable methods, but this requires us to unnecessarily copy the data:

```rust
    pub(crate) fn from_schema_field(object: &SchemaObject) -> Result<Self, String> {
        let field_ids = object.unique_field_ids();
        let mut entity = Entity::new();

        for field in field_ids {
            let mut data = SchemaComponentData::new();
			// Copy the data in the C API layer.
            data.fields_mut().copy_from(object.get_object(field))?;
            unsafe { entity.add_serialized(field, data)? };
        }

        Ok(entity)
    }

	pub(crate) unsafe fn add_serialized(&mut self, component_id: ComponentId, mut component: Owned<SchemaComponentData>) -> Result<(), String> {
        let vtable = DATABASE.get_vtable(component_id).unwrap();
        let deserialize_func = vtable.component_data_deserialize.unwrap_or_else(|| {
            panic!(
                "No component_data_deserialize method defined for {}",
                component_id
            )
        });
		
		// Proceed to deserialize the `Owned<SchemaComponentData>`
	}
```

So instead we type-erase the implementations of the `ObjectField` trait methods, push them into the vtable, and use those directly which means we get to avoid the copies.

spatialos-sdk/src/worker/vtable.rs
```diff
@@ -1,5 +1,96 @@
...
+        self.vtables.len()
+    }
+}
+
+unsafe impl Sync for ComponentDatabase {}
+unsafe impl Send for ComponentDatabase {}
+
+#[derive(Clone, Debug)]
+pub struct VTable {
+    pub worker_vtable: Worker_ComponentVtable,
```
randomPoison: Could this be `pub(crate)` instead of `pub`? It shouldn't need to be exposed publicly, as far as I can tell.

```suggestion
    pub(crate) worker_vtable: Worker_ComponentVtable,
```

jamiebrynes7: Done! :)

PR #143, opened 2020-01-04
`cargo-spatial` error handling consolidation

jamiebrynes7:
Following on from #139, this PR implements the same approach for the rest of the commands. In doing so, we generalise the error struct into its own module.

### DIFF
--- a/cargo-spatial/src/errors.rs
+++ b/cargo-spatial/src/errors.rs
@@ -1,15 +1,14 @@
 use serde::export::Formatter;
-use std::error::Error;
 use std::fmt::{Debug, Display};
 
 #[derive(Debug)]
-pub struct WrappedError<T: Display + Debug> {
+pub struct Error<T: Display + Debug> {
     pub kind: T,
     pub msg: String,
-    pub inner: Option<Box<dyn Error>>,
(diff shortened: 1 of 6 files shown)


cargo-spatial/src/errors.rs
```diff
@@ -0,0 +1,28 @@
+use serde::export::Formatter;
+use std::error::Error;
+use std::fmt::{Debug, Display};
+
+#[derive(Debug)]
+pub struct WrappedError<T: Display + Debug> {
+    pub kind: T,
+    pub msg: String,
+    pub inner: Option<Box<dyn Error>>,
+}
```
randomPoison: I kind of like this approach of having a generic struct plus a more context-specific enum 😁

It might be better to name it `Error`, though, instead of `WrappedError`. The general pattern for error types for rust is an `Error` struct plus an `ErrorKind` enum. In this case, you've made the `Error` struct generic over a couple different `ErrorKind` enums, but it's still effectively playing the role as the `Error` struct. It's not really a "wrapper" from the user's perspective.

jamiebrynes7: Yeah that makes sense :)

PR #144, opened 2020-01-05
Add schema `Entity` type

jamiebrynes7:
This PR adds the schema `entity` type. This corresponds to the `worker::entity::Entity` struct in Rust.

This schema type is structured like:

```
entity field
	field id 54 --> SchemaObject (position_component)
	field id 100 -> SchemaObject (some_component)
	...
```

such that a component is de/serialized from/to the field ID that corresponds to that component ID. We expose two new methods in the `VTable` struct to accommodate this structure which allow us to (de)serialize directly to/from the `SchemaObject`.

---

This PR also has a assorted set of changes to support this schema type.

- The `Component` trait requires `Clone` such that we can deep copy the `Entity` object. (The worker vtable "copy" is only a shallow copy). 
- Some convenience methods for mutating the `Entity` object.
- Exposing `unique_field_ids` on the `SchemaObject` type

### DIFF
--- a/spatialos-sdk/src/worker/entity.rs
+++ b/spatialos-sdk/src/worker/entity.rs
@@ -36,6 +36,35 @@ impl Entity {
         Ok(entity)
     }
 
+    pub(crate) fn from_schema_field(object: &SchemaObject) -> Result<Self, schema::Error> {
+        let field_ids = object.unique_field_ids();
+        let mut entity = Entity::new();
+
+        for field_id in field_ids {
+            entity
+                .pre_add_check(field_id)
+                .map_err(Error::schema_error::<Self>)?;
+
+            let component = object.get_object(field_id);
+            let mut owned = SchemaComponentData::new();
+            owned
+                .fields_mut()
+                .copy_from(component)
+                .map_err(Error::schema_error::<Self>)?;
+
+            entity.components.insert(field_id, owned);
+        }
+
+        Ok(entity)
+    }
+
+    pub(crate) fn to_schema_field(&self, object: &mut SchemaObject) {
+        for (component_id, data) in &self.components {
+            let component_field = object.add_object(*component_id);
+            component_field.copy_from(data.fields()).unwrap(); // TODO: Get rid of unwrap.
+        }
+    }
+
     pub(crate) fn add<C: Component>(&mut self, component: &C) -> Result<(), String> {
         self.pre_add_check(C::ID)?;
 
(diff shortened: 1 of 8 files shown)


jamiebrynes7:
I'll re-do this PR on top of #146 soon ™️

spatialos-sdk/src/worker/entity.rs
```diff
@@ -36,6 +36,35 @@ impl Entity {
...
+            entity.components.insert(field_id, owned);
+        }
+
+        Ok(entity)
+    }
+
+    pub(crate) fn to_schema_field(&self, object: &mut SchemaObject) {
+        for (component_id, data) in &self.components {
+            let component_field = object.add_object(*component_id);
+            component_field.copy_from(data.fields()).unwrap(); // TODO: Get rid of unwrap.
```
jamiebrynes7: @dgavedissian @randomPoison 

How do we want to handle the result from `SchemaObject::copy_from`? 

The underlying schema operations are fallible, although the only possible error (currently) is that the Schema Object is greater than 2 GB in size.

randomPoison: Huh, I'm surprised that the logic for copying a schema object is fallible when the logic for serializing into a schema object isn't 🤔 Is there anything wrong with returning a `Result` from `to_schema_field`?

jamiebrynes7: We can return a result from `to_schema_field` but it will simply be unwrapped/ignored at some point as serialization isn't fallible. 

> I'm surprised that the logic for copying a schema object is fallible when the logic for serializing into a schema object isn't 

Yeah this was confusing to me as well

randomPoison: If we need to we can totally make serialization fallible. On the other hand, there's probably no way to meaningfully recover from being unable to allocate enough space for your schema objects. It's pretty similar to an out-of-memory error, which Rust generally handles by panicking. So maybe it'd be fine to panic if the schema data can't be copied?

jamiebrynes7: Yeah that's sort of my opinion too. I'm kind of okay with leaving in this `unwrap()`

randomPoison: Maybe switch it to `expect` so that we can at least give some context on what's breaking when someone inevitably hits it 😛 Otherwise I'm good with having it panic 👍

jamiebrynes7: Also then satisfies the `TODO` comment :rocket:

randomPoison: Hah, true! 🎉

PR #145, opened 2020-01-09
Capture errors in vtable deserialization

randomPoison:
> Part of #121 

Updates vtable deserialization to capture errors and forward them to the user for handling. This allows user code to include special handling for the case that another worker is sending invalid schema data.

To make this change, I have moved the "user handle" abstraction into a dedicated `handle` module, and made some tweaks to how we represent user handles. Specifically:

* A user handle now always contains a `schema::Result<T>`, rather than a raw `T`. This is necessary because the vtable functions need to be written such that they assume they are receiving a `Result<T>`, so all places that create a user handle must provide a `Result<T>`, even if error handling isn't strictly necessary in this case. This invariant is enforced by the `handle::new` and `handle::new_raw` functions.
* Handles are now represented as an `Arc<dyn Any>`. This allows us to more cleanly represent untyped data in Rust-land without having to resort to raw pointers. `Entity` has (not yet) been updated to take advantage of this.


randomPoison:
After some discussion, we've decided to move away from using vtables entirely, so this PR is no longer relevant.

PR #146, opened 2020-01-13
Goodbye vtables

randomPoison:
This PR removes support for vtable serialization.

In doing so I also found a bug with how strings and bytes are handled during serialization. We were directly calling `Schema_AddBytes`, which doesn't actually copy the bytes data. We hadn't run into issues with this before because storing the component data in a user handle effectively addressed the lifetime issues, but once I removed vtable support some of the serialization tests started failing because strings were getting corrupted. I updated the serialization logic to allocate a separate buffer to copy the raw bytes into, similar to how adding primitive lists is handled.

Additional changes:

* Implemented `Clone` for owned schema data types (i.e. `Owned<T>`), and made it possible to clone a borrowed schema data object into an owned one (i.e. going from `&T` to `Owned<T>`).
* Simplified some of the test cases and improved the error messages.
* Re-implemented `Entity` to directly use `Owned<SchemaComponentData>` to store pre-serialized component data. This, in addition to removing vtables, removed the need for any unsafe code in `Entity`.
* Removed `RawEntity` and changed `Entity::raw_component_data` to return a `Vec<Worker_ComponentData>` directly.


spatialos-sdk/src/worker/connection.rs
```diff
@@ -424,8 +425,10 @@ impl Connection for WorkerConnection {
             reserved: ptr::null_mut(),
             component_id: C::ID,
             command_index,
-            schema_type: ptr::null_mut(),
-            user_handle: vtable::handle_allocate(request),
+            // TODO: Do we need to free the `Schema_CommandRequest` object at this point? Or will
+            // the SDK handle that for us?
```
jamiebrynes7: I believe the SDK handles that for you

spatialos-sdk/src/worker/entity.rs
```diff
@@ -26,114 +25,56 @@ impl Entity {
...
-    }
+            .iter_mut()
+            .map(|(&component_id, data)| Worker_ComponentData {
+                component_id,
 
-    pub(crate) fn raw_component_data(&self) -> RawEntity {
-        RawEntity::new(self.components.values())
+                // TODO: Why does this require a `*mut Schema_ComponentData`? Is there any actual
+                // chance that the underlying data will be mutated? Would it be safe for us to use
+                // `as_ptr` and cast it to a `*mut`?
```
jamiebrynes7: cc @dgavedissian ?

test-suite/src/snapshot_integration_tests.rs
```diff
@@ -22,42 +22,49 @@ pub fn writing_invalid_entity_returns_error() {
...
-            .expect("Error");
+            .expect("Failed to create `SnapshotOutputStream`")
+            .write_entity(EntityId::new(1), &mut entity)
+            .expect("Failed to write entity to snapshot");
     }
 
     {
-        let mut snapshot = SnapshotInputStream::new(snapshot_path).expect("Error");
+        let mut snapshot = SnapshotInputStream::new(snapshot_path)
+            .expect("Failed to create `SnapshotInputStream`");
```
jamiebrynes7: Honestly embarrassed by these previous "error" messages 😬

randomPoison: Hey, I'm just happy that you wrote the tests in the first place! I wouldn't have caught the error if we didn't have some tests in place 😁

spatialos-sdk/src/worker/connection.rs
randomPoison: Excellent, that's what I had hoped! I've removed those comments.

dgavedissian:
Planning on reviewing this sometime today

spatialos-sdk/src/worker/entity.rs
dgavedissian: Sorry, I'm not sure what this comment is referring to exactly?

randomPoison: For a little bit better context:

```rust
Worker_ComponentData {
    component_id,

    // TODO: Why does this require a `*mut Schema_ComponentData`? Is there any actual
    // chance that the underlying data will be mutated? Would it be safe for us to use
    // `as_ptr` and cast it to a `*mut`?
    schema_type: data.as_ptr_mut(),

    ..Default::default()
}
```

`Worker_ComponentData::schema_type` is a `*mut Schema_ComponentData`, which means that `Entity::raw_component_data` needs to take `&mut self`. This seems odd, since in practice it doesn't seem like there's any reason why that data would actually be mutated. My assumption is that `Worker_ComponentData` is also used in cases where the value of `schema_type` *will* be mutated (e.g. vtable serialization), so the pointer has to be a `*mut`.

In practice, would it be safe to instead do `data.as_ptr() as *mut _`? For that to be the case, the SDK would have to not actually mutate the schema data. I thin the functions where we're passing this data are `Worker_Connection_SendCreateEntityRequest` and `Worker_SnapshotOutputStream_WriteEntity`.

randomPoison: The more I think about it, the more confident that I feel that it should be safe to do `data.as_ptr() as *mut _`. I'm going to make that change now, but I'll hold off on merging until @dgavedissian can confirm.

dgavedissian: Note that in general, if you pass a `Schema_Whatever*` to a C API function, that function is actually going to take ownership over that data and free it when it's done with it. I haven't looked at the API in a while, but I believe that's the reason it's a mutable pointer.

See the following:
```
/**
 * Requests SpatialOS to create an entity. If components[i].schema_type is set, ownership is
 * transferred to the SDK, and components[i].schema_type is set to NULL. If
 * components[i].schema_type is NULL and components[i].user_handle is set, the entity data is
 * serialized immediately using the corresponding vtable serialize function.
 *
 * Returns Worker_RequestId -1 if the component is not registered and the default vtable is
 * not found.
 */
WORKER_API Worker_RequestId Worker_Connection_SendCreateEntityRequest(
    Worker_Connection* connection, uint32_t component_count, Worker_ComponentData* components,
    const Worker_EntityId* entity_id, const uint32_t* timeout_millis);
```

spatialos-sdk/src/worker/entity.rs
```diff
@@ -26,114 +25,70 @@ impl Entity {
...
+    ///
+    /// # Safety
+    ///
+    /// * The returned `Worker_ComponentData` objects borrow the serialized schema data from the
+    ///   `Entity`, and so cannot be used safely once the `Entity` is dropped.
+    /// * The returned `Worker_ComponentData` objects should only be used with C API functions that
+    ///   will not mutate the objects pointed to by the `schema_type` fields. In particular, this
+    ///   should be safe to use with `Worker_Connection_SendCreateEntityRequest` and
+    ///   `Worker_SnapshotOutputStream_WriteEntity`.
+    pub(crate) fn raw_component_data(&self) -> Vec<Worker_ComponentData> {
```
dgavedissian: Actually, you'll probably want to make this function bind to `self` rather than `&self` and relinquish ownership of the data, as that is the documented semantics for `Schema_*` objects (which is actually different to vtable based user handles - sorry).

dgavedissian: To preserve the existing behaviour, you could also add a `clone` function to `Entity` if it doesn't exist already after doing the change above, then change:
```
entity.raw_component_data()
```
to
```
entity.clone().raw_component_data()
```
in places where we call the C API.

jamiebrynes7: Alternatively, `raw_component_data()` could deep copy the schema data since there are built in schema functions for those sorts of things

spatialos-sdk/src/worker/entity.rs
randomPoison: Ah, excellent! I was looking for comments along that line explaining what would happening to `schema_type`, but I was looking in the wrong place (looking in the struct definition rather than the the function definition).

spatialos-sdk/src/worker/entity.rs
randomPoison: I like the plan of changing it to take `self` and implement `Clone` (which I _think_ I already did). That feels more idiomatic to me, in that you only pay the cost of deep-copying the data if you need to.

randomPoison: I've renamed the function to `into_raw()` and updated the comments to reflect the nuances of the new API 😎

spatialos-sdk/src/worker/entity.rs
randomPoison: Per your suggestion below, this comment is no longer necessary as we're now using `into_raw()` on the `SchemaComponentData` to get a `*mut Schema_ComponentData` :tada:

PR #148, opened 2020-01-18
Expose Schema API 

jamiebrynes7:
Simple case of just swapping a few fields/structs to `pub` 😄

### DIFF
--- a/.idea/runConfigurations/Check.xml
+++ b/.idea/runConfigurations/Check.xml
@@ -7,7 +7,9 @@
     <option name="emulateTerminal" value="false" />
     <option name="backtrace" value="SHORT" />
     <option name="workingDirectory" value="file://$PROJECT_DIR$" />
-    <envs />
+    <envs>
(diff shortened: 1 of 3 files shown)


.idea/runConfigurations/Check.xml
```diff
@@ -7,7 +7,9 @@
     <option name="emulateTerminal" value="false" />
     <option name="backtrace" value="SHORT" />
     <option name="workingDirectory" value="file://$PROJECT_DIR$" />
-    <envs />
+    <envs>
+      <env name="SPATIAL_LIB_DIR" value="./dependencies" />
```
dgavedissian: ?

jamiebrynes7: I missed setting this environment variable in #142

jamiebrynes7:
> Is this API possibly too low level / hard to use, or does it seem fine when used?

I think its good, not everyone will consume the API at this level, but we leave the option open. It's more or less the same level/complexity as the schema APIs in the C# bindings SDK.

PR #151, opened 2020-01-19
Remove generic type from `RequestId`

jamiebrynes7:
We used to do this in the C# SDK as well and its not really very helpful. 

Imagine you wanted to store a list of pending requests, you end up needing to type-erase the generic type in them. (I came across this when trying to build a demo and ended up falling back to the inner id)

It also makes the implementation a bit simpler.

### DIFF
--- a/spatialos-sdk/src/worker/mod.rs
+++ b/spatialos-sdk/src/worker/mod.rs
@@ -14,11 +14,8 @@ pub(crate) mod utils;
 pub mod worker_future;
 
 use component::ComponentId;
-use derivative::Derivative;
 use spatialos_sdk_sys::worker::Worker_InterestOverride;
-use std::cmp::Ordering;
 use std::fmt::{Display, Error, Formatter};
-use std::marker::PhantomData;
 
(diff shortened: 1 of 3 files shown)


spatialos-sdk/src/worker/mod.rs
```diff
@@ -45,45 +42,7 @@ impl Display for EntityId {
...
-
-impl<T> RequestId<T> {
-    pub fn new(id: i64) -> RequestId<T> {
-        RequestId {
-            id,
-            _type: PhantomData,
-        }
-    }
-}
+type RequestId = i64;
```
randomPoison: It might still be a good idea to make `RequestId` a newtype struct that wraps an `i64`. This would at least put a bit of a type-safe boundary around request IDs, such that users are less likely to accidentally provide an invalid value where a valid request ID is required.

jamiebrynes7: Done! `RequestId` looks similar to `EntityId` now :+1:

PR #153, opened 2020-02-02
Use exhaustive schema package

jamiebrynes7:
Opening this PR to track the various issues left in the code generator (as found by the exhaustive schema package).

- [x] Recursive schema options cause compiler errors (infinite size). These should be boxed. => #154 
- [x] Issues around using certain types in maps
	- Seems related to the constraint that the key in the map must implement `Ord` - https://github.com/jamiebrynes7/spatialos-sdk-rs/blob/tests/exhaustive-schema/spatialos-sdk/src/worker/schema/collections.rs#L151
	- Things like generated enums, generated types, schema entity, don't implement `Ord`.

### DIFF
--- a/.github/workflows/premerge.yml
+++ b/.github/workflows/premerge.yml
@@ -118,7 +118,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: spatial
-          args: --verbose download sdk --sdk-version 14.1.0
+          args: --verbose download sdk --sdk-version 14.1.0 --with-test-schema
         env:
           SPATIAL_LIB_DIR: "dependencies"
 
--- a/.github/workflows/premerge.yml
+++ b/.github/workflows/premerge.yml
@@ -36,7 +36,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: spatial
-          args: --verbose download sdk --sdk-version 14.1.0
+          args: --verbose download sdk --sdk-version 14.1.0 --with-test-schema
         env:
           SPATIAL_LIB_DIR: "dependencies"
 
--- a/.idea/runConfigurations/Download_SDK.xml
+++ b/.idea/runConfigurations/Download_SDK.xml
@@ -1,7 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Download SDK" type="CargoCommandRunConfiguration" factoryName="Cargo Command">
     <option name="channel" value="DEFAULT" />
-    <option name="command" value="spatial download sdk --sdk-version 14.1.0" />
+    <option name="command" value="spatial download sdk --sdk-version 14.1.0 --with-test-schema" />
     <option name="allFeatures" value="false" />
     <option name="nocapture" value="false" />
     <option name="emulateTerminal" value="false" />
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ To develop in this repository you'll need:
 4. Download the C API dependencies.
    
    ```
-   $ cargo spatial download sdk --sdk-version 14.1.0
+   $ cargo spatial download sdk --sdk-version 14.1.0 --with-test-schema
    ```
 5. Build the `spatialos-sdk` crate.
    
--- a/project-example/src/generated.rs
+++ b/project-example/src/generated.rs
@@ -5,6 +5,8 @@
 #![allow(dead_code)]
 #![allow(non_camel_case_types)]
 #![allow(unused_mut)]
+#![allow(clippy::unreadable_literal)]
+#![allow(clippy::option_option)]
 
 use spatialos_sdk::worker::schema::*;
 use spatialos_sdk::worker::component::*;
--- a/project-example/src/generated.rs
+++ b/project-example/src/generated.rs
@@ -112,7 +112,7 @@ impl Into<u32> for TestEnum {
 impl_field_for_enum_field!(TestEnum);
 
 /* Types. */
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct CommandData {
     pub value: i32,
 }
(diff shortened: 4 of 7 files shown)


jamiebrynes7:
@randomPoison @dgavedissian was there a specific reason we chose `BTreeMap` instead of `HashMap` for the schema map implementations? 

The constraints for `HashMap` are only `Hash` and `Eq`. I'm unsure how we would implement `Ord` for all generated types and the `Entity` type.

randomPoison:
> Also is that even true? Schemalang doesn't provide incremental collection updates

Not sure about this one. I *think* @dgavedissian gave that explanation when I last asked why we're using `BTreeMap` 🤷‍♀

dgavedissian:
Yeah that explanation is correct, because the C API supports byte level compression of component updates: https://docs.improbable.io/reference/14.4/shared/worker-configuration/component-settings#how-delta-compression-works

If we use a hash map, then the serialisation order is undefined.

jamiebrynes7:
Okay the last stumbling block that I'm struggling to solve is that we cannot implement `Ord` for any and all generated types since `f32/f64` doesn't implement `Ord`. 

What we could do is:

- Detect if a given schema type has any floats in it (be it a `list`, `map`, `option` or any child type). 
- If it doesn't => derive `Ord` and `Eq` for that type.
- If a user tries to generate schema from a map with a "invalid" type we could either skip the field or throw an error.

Thoughts @dgavedissian @randomPoison

randomPoison:
For custom schema types, the code generator could produce a custom `Ord` impl instead of using `#[derive(Ord)]`. This could do mostly the same thing as `#[derive(Ord)]`, but could include custom logic for floating point types that impose a total ordering around `NaN` values (e.g. `NaN` is considered equal to `NaN` and less than all other values).

jamiebrynes7:
> This could do mostly the same thing as #[derive(Ord)], but could include custom logic for floating point types that impose a total ordering around NaN values (e.g. NaN is considered equal to NaN and less than all other values).

I like this idea :)

jamiebrynes7:
Alright this is turning into a bit of a 'mare. 

To ensure correctness, we need to manually implement `PartialEq, Eq, PartialOrd, Ord` for the generated types to ensure that the implementations agree with each other. 

So I started implementing `PartialEq`, only special casing `f32/f64` fields and deferring to `self == other` for all other fields. All okay so far, but now I need to implement a custom impl for `Option<f32/f64>` since I can't use the built in for that either, and I'll need to do the same for `Vec<>` and `BTreeMap<f32/f64, f32/f64>`. 

And then do the same for `PartialOrd` and `Ord` 😬

randomPoison:
How about defining a custom `Ord` trait instead? We could use a blanket impl for all types that already implement `Ord`, and manually implement the trait for floating point types to enforce our own ordering:

```rust
pub trait FloatOrd: PartialOrd {
    float_cmp(&self, other: &Self) -> Ordering;
}

impl<T> FloatOrd for T where T: Ord {
    float_cmp(&self, other: &Self) -> Ordering {
        Ord::cmp(self, other)
    }
}

impl FloatOrd for f32 {
    float_cmp(&self, other: &Self) -> Ordering {
        self.partial_cmp(other).unwrap_or_else(|| {
            if self.is_nan() {
                if other.is_nan() {
                    Ordering::Equal
                } else {
                    Ordering::Less
                }
            } else {
                Ordering::Greater
            }
        })
    }
}

// TODO: Implement `FloatOrd` for all other basic schema types that don't implement `Ord`.
```

The generated code for all types would then be the exact same boilerplate of calling `FloatOrd::float_cmp` on each field.

jamiebrynes7:
We would then need our own `BTreeMap` equivalent that used `FloatOrd` instead of `Ord` that we would use for generated types. This just feels a little funky 😕

randomPoison:
We wouldn't need a custom `BTreeMap` type, since we could blanket impl `Ord` for all of our types implementing our custom `FloatOrd` trait.

jamiebrynes7:
Wouldn't that then mean there are competing implementations for Ord? Unless we can only do the blanket impl for types which don't already implement Ord

randomPoison:
Oh yeah, you're right about that. [You can't do blanket impls for an external trait](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=24d798580eedb5776ce7fe119b040626) 🤔

....

Oh! Okay, I think we can still make this work, though. We don't actually need a blanket impl, since we're already doing code generation. What we can do is still define a custom `FloatOrd` trait. We can do a blanket impl of `FloatOrd` for any type that already implements `Ord`, and manually implement it for the handful of built-in types that don't:

* `f32` and `f64`
* `Option<f32/f64>`
* `Vec<f32/f64>`
* `BTreeMap<f32/f64, ...>`

Then, in the generated `Ord` impl for any schema types, instead of deferring to the `Ord` impl for each of the fields, we defer to the `FloatOrd` impl! That should keep the generated code simple (i.e. the same boilerplate for each field) and minimize the amount of extra hassle needed.

jamiebrynes7:
Okay so the final nail in the coffin for the `FloatOrd` approach is that we cannot use floats in the key of a `BTreeMap` at all ([repro](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c312d7b55479de1a42aa2a606e300f27)) as the key has an `Ord` trait bound.

The other option on the table is to wrap the `f32/f64` in a newtype which does implement the required traits and use those when generating code. 

I.e. - something like https://github.com/notriddle/rust-float-ord

This has the cost of user ergonomics since the following is a compiler error:

```rust
let mut f = FloatOrd(0.5f32);
f = 1.0f32; // Cannot convert between f32 and FloatOrd<f32>
f.0 = 1.0f32 // The correct way of changing the floats
``` 

We can aid ergonomics by implementing `From<f32/64>` on the newtype, but its still not _great_. But it also feels like the only real way forward that I can see right now

jamiebrynes7:
Pushed a possible implementation of a `FloatOrd` newtype here https://github.com/jamiebrynes7/spatialos-sdk-rs/pull/165

jamiebrynes7:
Alright I _think_ this should be good to go now!

PR #154, opened 2020-02-02
Add recursive schema option support 

jamiebrynes7:
This PR introduces recursive schema option support. This is an edge case in schemalang where the following is valid schema.

```
type Recursive {
	option<Recursive> opt = 1;
}

type TypeA {
	option<TypeB> opt = 1;
}

type TypeB {
	option<TypeA> opt = 1;
}
```

To support this edge case, we detect when a field is recursive (i.e. - cyclical) and change the type from `Option<T>` to `Option<Box<T>>`. 

To enable this, we add another schema container `RecursiveOptional<T>` which is almost identical to `Optional<T>` except it has an additional trait bound `ObjectField` since only types can be optional _and_ the `RustType` alias has the additional `Box`. 

---

For the above schema this change should now generate:

```rust
struct Recursive {
	pub opt: Option<Box<Recursive>>
}

struct TypeA {
	pub opt: Option<Box<TypeB>>
}

struct TypeB {
	pub opt: Option<Box<TypeA>>
}
```

### DIFF
--- a/project-example/schema/example.schema
+++ b/project-example/schema/example.schema
@@ -60,3 +60,8 @@ component EntityTest {
 
     Entity entity = 1;
 }
+
+// TODO: Remove when https://github.com/jamiebrynes7/spatialos-sdk-rs/pull/153 is merged.
+type Recursive {
+    option<Recursive> opt = 1;
+}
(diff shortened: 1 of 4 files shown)


randomPoison:
Are options the only place where we need to support recursive types? I like not having to box the inner values if the type isn't recursive, but I worry that that's going to be difficult to support robustly if arbitrary types can be recursive.

jamiebrynes7:
Yup options are the only place where we have to special case them. 😄

PR #161, opened 2020-02-18
Add codecov config file

jamiebrynes7:
Added a very minimal codecov configuration which shouldn't fail PRs any more regarding code coverage, but it should still be reported.


randomPoison:
Glad PRs won't be blocked on code coverage 🎉

jamiebrynes7:
They were never blocked, but the red cross made me sad (maybe that's the point 🤔)

PR #163, opened 2020-03-11
Primitive schema type defaults

Chris-Nicholls:
Added defaults when deserialising primitive types, to match pbio as seen in https://github.com/improbable/protobuf-interop/blob/4f1363cf2522dc97eba7b669c00731a88b2de9e4/src/pbio.cc#L367

For some reason, default values are not present when sent from spatial, leading to errors when deseralising things such as Position(0,0,0)

### DIFF
--- a/spatialos-sdk/src/worker/schema/primitives.rs
+++ b/spatialos-sdk/src/worker/schema/primitives.rs
@@ -34,6 +34,7 @@ macro_rules! impl_primitive_field {
         $schema_add:ident,
         $schema_add_list:ident,
         $schema_get_list:ident,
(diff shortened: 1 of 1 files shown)


project-example/world.json
```diff
@@ -1,5 +1,5 @@
 {
-  "template": "small",
+  "template": "w2_r0500_e5",
```
Chris-Nicholls: I'm surprised that github doesn't pick up that this is already  merged in

jamiebrynes7: Possibly related to that this branch is out of date?

PR #157, opened 2020-02-15
Upgrade to Worker SDK 14.1.0

jamiebrynes7:
This PR contains (most of) the upgrade to the Worker SDK 14.1.0. 

It notably is missing:

- The `Schema_Bundle` and `Schema_Json` APIs
- JSON snapshot support

The first of these is implemented, but I couldn't quite get it to work, so I figured lets get the rest of this in.

PR #164, opened 2020-03-19
Add schema bundles

jamiebrynes7:
The last bit of the 14.1.0 upgrade that I ommitted in #157. This PR adds support for schema bundles. 

There are a number of tests except for the command conversions. I used the schema standard library to generate the binary bundle representations which doesn't have any interesting commands (only empty ones).

In a follow up PR, I'm going to add an option to generate a binary bundle as well as the JSON one in `cargo spatial codegen` and use that to generate more interesting bundles. This is just to keep the PR a bit smaller (its already fairly large).

### DIFF
--- a/spatialos-sdk/src/worker/schema/bundle.rs
+++ b/spatialos-sdk/src/worker/schema/bundle.rs
@@ -1,0 +1,321 @@
+use crate::worker::{component::ComponentId, schema::*, utils::cstr_to_string};
+use spatialos_sdk_sys::worker::*;
+use std::ffi::CString;
+use std::ptr::NonNull;
+
+pub type JsonConversionResult<T> = std::result::Result<(T, Option<String>), String>;
+
+pub struct Bundle {
+    ptr: NonNull<Schema_Bundle>,
+}
+
+impl Bundle {
+    pub fn load(buffer: &[u8]) -> std::result::Result<Self, String> {
+        unsafe {
+            let ptr = Schema_Bundle_Load(buffer.as_ptr(), buffer.len() as u32);
+            let ptr = NonNull::new(ptr)
+                .ok_or_else(|| "Received null pointer from Schema_Bundle_Load".to_string())?;
+            let err = Schema_Bundle_GetError(ptr.as_ptr());
+
+            if !err.is_null() {
+                let msg = cstr_to_string(err);
+                Schema_Bundle_Destroy(ptr.as_ptr());
+                Err(msg)
+            } else {
+                Ok(Bundle { ptr })
+            }
+        }
+    }
+
+    pub fn load_object<T: AsRef<str>, U: AsRef<str>>(
+        &self,
+        qualified_type_name: T,
+        json: U,
+        dest: &mut SchemaObject,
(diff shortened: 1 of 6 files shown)


spatialos-sdk/src/worker/schema/bundle.rs
```diff
@@ -0,0 +1,620 @@
...
+            Ok(((), Bundle::get_last_warning()))
+        }
+    }
+
+    // TODO: Why is this mut?
+    pub fn dump_object<T: AsRef<str>>(
+        &self,
+        qualified_type_name: T,
+        src: &mut SchemaObject,
+    ) -> JsonConversionResult<String> {
```
randomPoison: What's this TODO referring to? Is this something that needs to be addressed?

spatialos-sdk/src/worker/schema/bundle.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    unsafe fn load_generic<
+        T: OwnedPointer + ToOwned<Owned = Owned<T>>,
+        F: FnOnce() -> *mut T::Raw,
+    >(
+        &self,
+        load: F,
+    ) -> JsonConversionResult<Owned<T>> {
+        let data = load();
```
randomPoison: What's the reason for `load` being a function? If it's a `FnOnce` that doesn't take an parameters and is called immediately within `load_generic`, wouldn't it be the same to have the code calling `load_generic` directly pass the result of the callback into `load_generic`, instead of passing in an `FnOnce`? The same goes for `dump_generic` below.

spatialos-sdk/src/worker/schema/bundle.rs
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    #[test]
+    pub fn loading_bundle_succeeds_when_valid_bundle() {
+        assert!(read_bundle(true).is_ok(), "Valid bundle failed to load.")
+    }
+
+    // TODO: This test currently fails (but it shouldn't!)
+    //#[test]
+    pub fn loading_bundle_fails_when_invalid_bundle() {
```
randomPoison: Did you want to get this test working as part of this PR? Or will that need to wait for a future update?

spatialos-sdk/src/worker/schema/bundle.rs
jamiebrynes7: It was something for me to investigate. I think its along the same lines of the interior mutability discussion we had a little while ago but need to confirm.

spatialos-sdk/src/worker/schema/bundle.rs
jamiebrynes7: Yeah this seems like some overzealous refactoring. Will simplify as suggested :)

spatialos-sdk/src/worker/schema/bundle.rs
jamiebrynes7: A future update, this test revealed a bug in the underlying implementation. The fix was released in `14.5.0` so I'll update this TODO.

PR #165, opened 2020-03-22
`FloatOrd` newtype

jamiebrynes7:
Possible implementation for https://github.com/jamiebrynes7/spatialos-sdk-rs/pull/153#issuecomment-602252163 wherein we try to use a wrapper newtype to make `f32/f64` implement `Ord`. 

This code was adapted from https://github.com/notriddle/rust-float-ord with a set of trait implementations added to aid usability: `Deref`, `DerefMut`, `+-/*`, and the assign variants thereof.


randomPoison:
As unfortunate as it is to not be able to use `f32`/`f64` directly, this is probably the most reasonable solution for now. I'm in favor of getting this merged in 👍

PR #166, opened 2020-04-11
Command traits

jamiebrynes7:
Following on from @randomPoison's ideas in #121 and examples in various previous PRs, this PR splits out  the `CommandResponse` and `CommandRequest` traits from the `Component` trait. 

It also restructures the traits somewhat such that there is a `Commands` trait which is only implemented if there any commands associated with that component. This cuts down on unnecessary generated code and models the domain (schema) a bit more accurately.

There are a number of associated changes in this PR:

- Refactored `StatusCode` in command response ops to just be a plain old `Result<T, E>`.
- Removed some unused types/methods
- Moved command related things into the `commands.rs` file

### DIFF
--- a/spatialos-sdk/src/worker/commands.rs
+++ b/spatialos-sdk/src/worker/commands.rs
@@ -1,3 +1,4 @@
+use crate::worker::entity::Entity;
 use crate::worker::query::EntityQuery;
 use crate::worker::EntityId;
 use spatialos_sdk_sys::worker::Worker_CommandParameters;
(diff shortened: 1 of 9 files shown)


spatialos-sdk/src/worker/commands.rs
```diff
@@ -1,6 +1,136 @@
...
+            command_index: request.command_index,
+            schema_type: SchemaCommandRequest::from_raw(request.schema_type),
+        }
+    }
+
+    // NOTE: We manually declare that the request impl `ObjectField`
+    // here, but in practice this will always be true for all component types. Future
+    // iterations should clean this up such that the `Component` trait can imply these
+    // other bounds automatically (i.e. by making them super traits of `Component`).
+    pub(crate) fn get<C: Commands>(&self) -> Option<schema::Result<C::Request>> {
```
randomPoison, line 91: I think this comment is no longer relevant now that the `Commands` trait guarantees that the command can be serialized.

spatialos-sdk/src/worker/commands.rs
```diff
@@ -1,6 +1,136 @@
...
+            command_index: response.command_index,
+            schema_type: SchemaCommandResponse::from_raw(response.schema_type),
+        }
+    }
+
+    // NOTE: We manually declare that the response impl `ObjectField`
+    // here, but in practice this will always be true for all component types. Future
+    // iterations should clean this up such that the `Component` trait can imply these
+    // other bounds automatically (i.e. by making them super traits of `Component`).
+    pub fn get<C: Commands>(&self) -> Option<schema::Result<C::Response>> {
```
randomPoison, line 123: Same goes for this comment, can be removed now.

jamiebrynes7:
> Sorry for taking so long to get to this, btw!

Oh no, don't worry about it 😄

PR #167, opened 2020-04-28
[WIP] downloading dependencies at build time

Chris-Nicholls:
Requiring the environment variable "SPATIAL_LIBS_DIR" has been causing me a bunch of issues (cargo check can't be run by IDEs for example).

This change has helped me with my setup significantly - it downloads the dependencies when spatial-os-sdk is built, and links to them in that directory. It also doens't redownload every time which significantly helps compile times. 

Not comprehensive, but wanted to see what you thought of doing it this way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Cargo.lock
 # SpatialOS SDK dependencies.
 dependencies/
 .idea/workspace.xml
+/spatialos-sdk-sys/dependecies/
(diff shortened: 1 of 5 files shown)


jamiebrynes7:
I think we could probably merge the two approaches? 

First look for a lib in `SPATIAL_LIB_DIR` (or possibly specified by `Spatial.toml`?), then try to download the libraries. 

This way you get the benefit of:

- IDE is happier
- less friction when doing builds for the first time
- ability to build in a clean environment by controlling dependencies fully (i.e. - building in a Docker image, I don't want to have to add `spatial`, deal with auth, etc). 

There's also scope for being more intelligent about partial downloads for the platform you are building for (rather than _everything_). 

Thoughts @Chris-Nicholls?

jamiebrynes7:
Merged #168 for downloading the packages at build time if `SPATIAL_LIB_DIR` isn't set.

PR #169, opened 2020-05-08
Make patch reporting never fail

jamiebrynes7:
Really just want this for informational purposes


jamiebrynes7:
Yeah, it wasn't blocking PRs, but it was marking the check as failed which was sad 😢

